PR #267, opened 2021-06-08
Implement missing user backend methods

juliushaertl:
Implements #7

### DIFF
--- a/lib/User/Backend.php
+++ b/lib/User/Backend.php
@@ -28,17 +28,22 @@ namespace OCA\UserOIDC\User;
 use OCA\UserOIDC\AppInfo\Application;
 use OCA\UserOIDC\Db\UserMapper;
 use OCP\AppFramework\Db\DoesNotExistException;
+use OCP\DB\Exception;
 use OCP\User\Backend\ABackend;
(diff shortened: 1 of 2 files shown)


ChristophWurst (approved):

lib/User/Backend.php
```diff
@@ -28,29 +28,44 @@
...
+			$this->logger->error('Failed to delete user', [ 'exception' => $e ]);
+			return false;
+		}
 	}
 
 	public function getUsers($search = '', $limit = null, $offset = null) {
-		return [];
+		return array_map(function($user) {
+			return $user->getUserId();
+		},$this->userMapper->find($search, $limit, $offset));
```
eneiluj, line 67: :grin:
```suggestion
		}, $this->userMapper->find($search, $limit, $offset));
```

eneiluj (changes requested):
All good except a very important syntax detail.

lib/User/Backend.php
juliushaertl: Fixed :)

eneiluj (approved):

Issue #42, opened 2020-08-07
Token <--> Property mapping


gargamel-cddr:
I used the community openid connect application, which allowed me to define which token property gets mapped to the nextcloud username, quotaSettings, etc.

With this solution, I cannot do that anymore. This results in 
a) my users not being able to find their files and 
b) creation of new users that
c) cannot be found on the user page

I expect to make this mapping configurable, or at least honor the (preferred_)username property of a token. Right now, it seems to just use the *sub* of the token. And to be honest, this official user backend should at least provide the same functionality like the community plugin does.

rullzer:
Migration is not supported from one to the other.
You can still just use the community one if that works better for you.

gargamel-cddr:
Well, as a matter of fact, I cannot do so, because this Plugin doesn‘t uninstall properly, but that‘s another issue I need to open.
But Migration is not my main point here: sub + issuer should be used for uniqueness of the account, but not the username. 

In my opinion, preferred_username from the openid Profile should be used for this. Also, this Plugin does not Support Quota Settings and the Users cannot be found in the User Store. All this is being done by the Community Plugin, therefore I mentioned it so you can get „inspired“ by their implementation. It provides nice Features, and I would‘ve sticked to it if it wasn‘t broken in latest times.

juliushaertl:
That should be possible now in case the OIDC token contains a field for the old user id in use with disabling the unique uid option and setting a custom mapping for the uid with #268

PR #268, opened 2021-06-08
Custom attribute mappings

juliushaertl:
- Allow mapping of uid/displayName/email/quota
- Add option to turn off the unique hasing of user ids
- Allow updating existing provider configurations
- Extend frontend for new settings and provider updating

based on #267 
implements #42

## Updated settings UI

image
image

### DIFF
--- a/lib/Settings/AdminSettings.php
+++ b/lib/Settings/AdminSettings.php
@@ -28,6 +28,7 @@ namespace OCA\UserOIDC\Settings;
 use OCA\UserOIDC\AppInfo\Application;
 use OCA\UserOIDC\Db\ProviderMapper;
 use OCA\UserOIDC\Service\ID4MeService;
+use OCA\UserOIDC\Service\ProviderService;
 use OCP\AppFramework\Http\TemplateResponse;
 use OCP\IInitialStateService;
 use OCP\IURLGenerator;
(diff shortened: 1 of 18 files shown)


lib/Settings/AdminSettings.php
```diff
@@ -37,7 +38,7 @@
 class AdminSettings implements ISettings {
 
 	/** @var ProviderMapper */
```
blizzz, line 40: 😕

package.json
```diff
@@ -24,6 +24,7 @@
 		"@nextcloud/initial-state": "^1.2.0",
 		"@nextcloud/logger": "^2.0.0",
 		"@nextcloud/router": "^2.0.0",
+		"@nextcloud/vue": "^4.0.0-alpha.4",
```
blizzz, line 27: alpha?

blizzz (commented):
Looks good, test might be a bit shallow. Comments inline.

blizzz (approved):
didn't test, looks good

juliushaertl:
Put some basic unit tests for the new logic except for the controller which should rather be covered by proper integration tests anyways. First attempts to fire up some presetup oidc provider didn't work well but I'll keep looking into that in a separate pr then.

PR #284, opened 2021-06-18
Add commandline setup for app

tsdicloud:
### Status
The OpenId app could only be configured by user interface.

### Proposal:
At least the provider configuration should be possible via commandline
to support automated operations.

>To show provider configuration, use:
>```
>sudo -u www-data php /ver/www/nextcloud/occ oidc:provider demoprovider
>```
>
>A provider is created if none with the given identifier exists and all parameters are given:
>```
>sudo -u www-data php /ver/www/nextcloud/occ oidc:provider demoprovider --clientid="WBXCa003871" \
>    --clientsecret="lbXy***********" --discoveryuri="https://accounts.central.de/openid-configuration"
>```
>
>To delete a provider, use:
>```
>sudo -u www-data php /ver/www/nextcloud/occ oidc:provider:remove demoprovider
> Are you sure you want to delete OpenID Provider demoprovider
> and may invalidate all assiciated user accounts.
>```
> To skip the confirmation, use `--force`.

The description of the commands as well as a link to a tutorial about the app usage is also included in the pull request.

### Limitations:
The new field mapping functionality is not backed up by commands (yet).

### DIFF
--- a/lib/Command/UpsertProvider.php
+++ b/lib/Command/UpsertProvider.php
@@ -27,6 +27,7 @@ use OCA\UserOIDC\Db\ProviderMapper;
 use OCA\UserOIDC\Db\Provider;
 
 use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 
@@ -54,7 +55,7 @@ class UpsertProvider extends Base {
 		$providerid   = $input->getArgument('providerid');
 		$clientid     = $input->getOption('clientid');
 		$clientsecret = $input->getOption('clientsecret');
-		$discoveriuri = $input->getOption('discoveryuri');
+		$discoveryuri = $input->getOption('discoveryuri');
 
 		if (false === $clientid) {
 			// in this case, the option was not passed when running the command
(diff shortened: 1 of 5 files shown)


lib/Command/UpsertProvider.php
```diff
@@ -0,0 +1,91 @@
...
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OCA\UserOIDC\Command;
+
+use OC\Core\Command\Base;
```
ChristophWurst, line 26: apps are not supposed to use `\OC\*` classes and if I'm not mistaken none of the functionality from that base class is actually used there. You can extend `\Symfony\Component\Console\Command\Command` directly.

lib/Command/UpsertProvider.php
```diff
@@ -0,0 +1,91 @@
...
+			->addOption('clientsecret', 's', InputOption::VALUE_REQUIRED, 'OpenID client secret')
+			->addOption('discoveryuri', 'd', InputOption::VALUE_REQUIRED, 'OpenID discovery endpoint uri');
+		parent::configure();
+	}
+
+	protected function execute(InputInterface $input, OutputInterface $output) {
+		$providerid = $input->getArgument('providerid');
+		$clientid = $input->getOption('clientid');
+		$clientsecret = $input->getOption('clientsecret');
+		$discoveryuri = $input->getOption('discoveryuri');
```
ChristophWurst, line 57: for type safety you might want to cast the four input arguments/options here to their respective types

lib/Db/ProviderMapper.php
```diff
@@ -62,4 +84,60 @@ public function getProviders() {
...
+			return $this->update($provider);
+		}
+	}
+
+	/**
+	 * Create or update provider settinngs
+	 *
+	 * @param string identifier
+	 */
+	public function deleteProvider(string $identifier) {
```
ChristophWurst, line 135:
```suggestion
	public function deleteProvider(string $identifier): void {
```

lib/Command/UpsertProvider.php
tsdicloud: This is not so easy because of the optionl handling of the fields on updates. Tried it out and it worked not so well. All types are strings and the createOrupdate signature is typed, so I would assume implicit casting is enough.

lib/Command/UpsertProvider.php
```diff
@@ -0,0 +1,91 @@
...
+	public function __construct(ProviderMapper $providerMapper) {
+		parent::__construct();
+		$this->providerMapper = $providerMapper;
+	}
+
+	protected function configure() {
+		$this
+			->setName('user_oidc:provider')
+			->setDescription('Create, show or update a OpenId connect provider connfig given the identifier of a provider')
+			->addArgument('providerid', InputOption::VALUE_REQUIRED, 'Administrative identifier name of the provider in the setup')
```
juliushaertl, line 46: The issue with using the provider identifier as of right now is that it is not enforced to be unique currently, so the frontend as well as the mapper allow creating two providers with the same one. I'll prepare a patch to add that separately, so we can get this merged and I can then extend it with the attribute mapping in that pull request.

lib/Command/UpsertProvider.php
```diff
@@ -0,0 +1,91 @@
...
+		parent::configure();
+	}
+
+	protected function execute(InputInterface $input, OutputInterface $output) {
+		$providerid = $input->getArgument('providerid');
+		$clientid = $input->getOption('clientid');
+		$clientsecret = $input->getOption('clientsecret');
+		$discoveryuri = $input->getOption('discoveryuri');
+
+		if (false === $clientid) {
```
juliushaertl, line 59: Would not be needed, if the option is not passed it will be null already.

juliushaertl:
@tsdicloud Mind if i take over doing some further adjustments and working on the attribute part right away?

tsdicloud:
> @tsdicloud Mind if i take over doing some further adjustments and working on the attribute part right away?

NO fine for me. Configuring the new provider is our task anyway for next week,  so we should be able to automate it later by commandline.

juliushaertl:
Rebased and continued in https://github.com/nextcloud/user_oidc/pull/292 as I couldn't push to the original branch.

PR #292, opened 2021-06-30
Add provider admin commands

juliushaertl:
Continuing the work from #284 with extension of the custom attribute mapping and some improvements to make sure the provider identifier is enforced to be unique.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,9 @@
   "require": {
     "coenjacobs/mozart": "^0.7.1",
     "id4me/id4me-rp": "^1.2",
-    "firebase/php-jwt": "^5.2"
+    "firebase/php-jwt": "^5.2",
+    "symfony/event-dispatcher": "^4",
+    "symfony/console": "^4"
   },
   "require-dev": {
     "nextcloud/coding-standard": "^0.5.0"
--- a/lib/Command/UpsertProvider.php
+++ b/lib/Command/UpsertProvider.php
@@ -23,10 +23,12 @@ declare(strict_types=1);
 
 namespace OCA\UserOIDC\Command;
 
+use Exception;
 use \Symfony\Component\Console\Command\Command;
 
 use OCA\UserOIDC\Db\ProviderMapper;
 
+use Symfony\Component\Console\Helper\Table;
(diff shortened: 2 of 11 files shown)


composer.json
```diff
@@ -23,7 +23,9 @@
   "require": {
     "coenjacobs/mozart": "^0.7.1",
     "id4me/id4me-rp": "^1.2",
-    "firebase/php-jwt": "^5.2"
+    "firebase/php-jwt": "^5.2",
+    "symfony/event-dispatcher": "^4",
+    "symfony/console": "^4"
```
ChristophWurst, line 28: is this for static analysis?

ChristophWurst (changes requested):

composer.json
juliushaertl: No it is being pulled in by `coenjacobs/mozart`, but I guess that could actually be a dev dependency only.

juliushaertl: Hm ok, but then we cannot built the autoloader anymore with out the dev depencenies :/

juliushaertl: Just found https://github.com/nextcloud/mail/pull/4633 let me prepare something similar for that here as well

juliushaertl: Made an independent PR at https://github.com/nextcloud/user_oidc/pull/296

lib/Command/UpsertProvider.php
```diff
@@ -0,0 +1,186 @@
...
+				$this->providerService->setSetting($provider->getId(), ProviderService::SETTING_MAPPING_EMAIL, $mapping);
+			}
+			if ($mapping = $input->getOption('mapping-quota')) {
+				$this->providerService->setSetting($provider->getId(), ProviderService::SETTING_MAPPING_QUOTA, $mapping);
+			}
+			if ($mapping = $input->getOption('mapping-uid')) {
+				$this->providerService->setSetting($provider->getId(), ProviderService::SETTING_MAPPING_UID, $mapping);
+			}
+		} catch (Exception $e) {
+			$output->writeln($e->getMessage());
```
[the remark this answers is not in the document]
juliushaertl: Adjusted accordingly

ChristophWurst (approved):

PR #293, opened 2021-06-30
Event for attribute mapping customzation

juliushaertl:
This allows to implement custom parsing of attributes before their value is applied during user provisioning.


juliushaertl:
Closing as it will be included in https://github.com/nextcloud/user_oidc/pull/317

PR #317, opened 2021-08-31
Claim handling and complex mapping rules

tsdicloud:
This pull request contains an extension that has been needed to integrate with a big, corporate identity management system.

## Part 1: Support OpenId connect claims
OpenID connect claims define a set of attributes that is passes together with the token back to the client that is requesting authorisation. A client can select the attribute sets in its request to IDM by listing attribute profile names in`scope` field 
The `scope` field is added to the Provider configuration and defaults to `openid email profile'

## Part 2: Complex mapping rules coded with events
Before a field in user account is filled with the mapped field in the token, the value of the token is published as `AttributeMappedEvent`. The event contains all claims as event payload, as well as the value of the mapped source claim in token. If the mapped claim does not exist in token, the `null` value is set as default in event.

The `AttributeMappedEvent::handle(Event $event)` can implement complex code for rules based on multiple claims and pass back the result by `setValue`. With using the event system, the rules can be implemented in a separate app which decouples business rule logic from `user_oidc` technical code.

If the value of an event is `null`and there is no directly mapped field value available, the field in user account is not set.

## Part 3: query parameters for discovered `token_endpoint`
For some developers, it is tempting to extend functionality of the IDM `token_endpoint` by introducing custom query parameter. This is not good for a standard OpenID connect client implementation, as the `token_endpoint` is called as a second step after `discovery` of the OPenId connect endpoints. To support such query parameters and not implement a `custom`case for it. the `discovery_url` can be configured in provider with query parameters. The query is send with the discover AND with the token endpoint. This usually works well because custom query parameter can usually be send to discovery endpoint but are ignored there. After discovery, the same query parameters are again send to `token_endpoint`and accepted there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,9 +76,17 @@ The app uses [krankerl](https://github.com/ChristophWurst/krankerl) to build the
 The release will be put into `build/artifacts/` when running the `krankerl package`.
 
 The app can also be built without krankerl by manually running:
-
 ```
 composer install --no-dev -o
-npm install
+npm install npm
 npm run build
 ```
+
+On Ubuntu 20.04, a possible way to get build working is with matching npm and node versions is:
+```
+sudo apt-get remove nodejs
+sudo curl -sL https://deb.nodesource.com/setup_16.x | sudo -E bash - 
+sudo apt-get install nodejs
+```
+
+
--- a/README.md
+++ b/README.md
@@ -78,15 +78,16 @@ The release will be put into `build/artifacts/` when running the `krankerl packa
 The app can also be built without krankerl by manually running:
 ```
 composer install --no-dev -o
-npm install npm
+npm ci
 npm run build
 ```
 
 On Ubuntu 20.04, a possible way to get build working is with matching npm and node versions is:
 ```
 sudo apt-get remove nodejs
-sudo curl -sL https://deb.nodesource.com/setup_16.x | sudo -E bash - 
+sudo curl -sL https://deb.nodesource.com/setup_14.x | sudo -E bash - 
 sudo apt-get install nodejs
+sudo npm install -g npm@7
 ```
 
 
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -3,8 +3,8 @@
   xsi:noNamespaceSchemaLocation="https://apps.nextcloud.com/schema/apps/info.xsd">
 	<id>user_oidc</id>
 	<name>OpenID Connect user backend</name>
-	<summary>Use an OpenID Connect backed to login to your Nextcloud</summary>
-	<description>Allows felxible configuration of an OIDC server as Nextcloud login user backend.</description>
+	<summary>Use an OpenID Connect backend to login to your Nextcloud</summary>
+	<description>Allows flexible configuration of an OIDC server as Nextcloud login user backend.</description>
 	<version>1.1.0-dev1</version>
 	<licence>agpl</licence>
 	<author>Roeland Jago Douma</author>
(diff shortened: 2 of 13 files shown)


juliushaertl:
> TypeError: Argument 5 passed to OCA\UserOIDC\Db\ProviderMapper::createOrUpdateProvider() must be of the type string, null given, called in /home/runner/work/user_oidc/user_oidc/apps/user_oidc/lib/Command/UpsertProvider.php on line 128 and defined in /home/runner/work/user_oidc/user_oidc/apps/user_oidc/lib/Db/ProviderMapper.php:99

```
 There was 1 failure:

1) ProviderServiceTest::testGetProvidersWithSettings
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
         'clientId' => null
         'discoveryEndpoint' => null
         'settings' => Array (...)
+        'scope' => null
     )
     1 => Array (
         'id' => 2
@@ @@
         'clientId' => null
         'discoveryEndpoint' => null
         'settings' => Array (...)
+        'scope' => null
     )
 )

/home/runner/work/user_oidc/user_oidc/apps/user_oidc/tests/unit/Service/ProviderServiceTest.php:98
```

juliushaertl (approved):

README.md
```diff
@@ -76,9 +76,17 @@ The app uses [krankerl](https://github.com/ChristophWurst/krankerl) to build the
 The release will be put into `build/artifacts/` when running the `krankerl package`.
 
 The app can also be built without krankerl by manually running:
-
 ```
 composer install --no-dev -o
-npm install
+npm install npm
```
eneiluj, line 81:
```suggestion
npm ci
```

appinfo/info.xml
```diff
@@ -3,11 +3,13 @@
   xsi:noNamespaceSchemaLocation="https://apps.nextcloud.com/schema/apps/info.xsd">
 	<id>user_oidc</id>
 	<name>OpenID Connect user backend</name>
-	<summary>Use an openID connect backed to login to your Nextcloud</summary>
-	<description>Allow configuring an OIDC server as Nextcloud user backend.</description>
-	<version>1.0.0</version>
+	<summary>Use an OpenID Connect backed to login to your Nextcloud</summary>
```
eneiluj, line 6:
```suggestion
	<summary>Use an OpenID Connect backend to login to your Nextcloud</summary>
```

appinfo/info.xml
```diff
@@ -3,11 +3,13 @@
...
+	<description>Allows felxible configuration of an OIDC server as Nextcloud login user backend.</description>
```
eneiluj, line 7:
```suggestion
	<description>Allows flexible configuration of an OIDC server as Nextcloud login user backend.</description>
```

lib/Db/Provider.php
```diff
@@ -36,6 +36,10 @@
  * @method void setClientSecret(string $clientSecret)
  * @method string getDiscoveryEndpoint()
  * @method void setDiscoveryEndpoint(string $discoveryEndpoint)
+ * @method string getScope()
+ * @method void setScope(string $scope)
+ * @method string getCustomQuery()
+ * @method void setCustomQuery(string $customQuery)
```
eneiluj, line 42:
```suggestion
```
I guess this was forgotten here.

src/components/SettingsForm.vue
```diff
@@ -53,6 +53,13 @@
 				type="text"
 				required>
 		</p>
+		<p>
+			<label for="oidc-scopr">{{ t('user_oidc', 'Scope') }}</label>
```
eneiluj, line 57:
```suggestion
			<label for="oidc-scope">{{ t('user_oidc', 'Scope') }}</label>
```

eneiluj (changes requested):
Looking good!
A few details to adjust and we're good to go.
Thanks a lot for this PR.

eneiluj (approved):

PR #320, opened 2021-09-06
Bearer token validation

juliushaertl:
Based on https://github.com/nextcloud/user_oidc/pull/317

- Move bearer token validation to separate classes
  - Self-contained access token (JWT)
  - UserInfo based validation
  - Possible for later: using token introspection endpoint
- Move some OpenID connect API calls to a dedicated service

## TODO

- [x] Make bearer token check optional as it has performance impact and maybe not be used on all setups (done in #321)

PR #321, opened 2021-09-06
make bearer check in API/WebDav calls optional

eneiluj:
This adds (in #320) an option to toggle Bearer token check for each provider. This new option is used in the user backend.

For now, this option value is stored the "old fashioned" way: in `oc_appconfig` instead of the `oc_user_oidc_providers` table (like it's been recently done for `scope`).
Would you prefer to store it in `oc_user_oidc_providers`?

It's been successfully tested on my test environment.

I'll make a review of #320 once this is merged (or not :grin:).

### DIFF
--- a/lib/User/Backend.php
+++ b/lib/User/Backend.php
@@ -25,6 +25,7 @@ declare(strict_types=1);
 
 namespace OCA\UserOIDC\User;
 
+use OCA\UserOIDC\Service\ProviderService;
 use OCA\UserOIDC\User\Validator\SelfEncodedValidator;
 use OCA\UserOIDC\User\Validator\UserInfoValidator;
 use OCA\UserOIDC\AppInfo\Application;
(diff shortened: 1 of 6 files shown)


lib/User/Backend.php
```diff
@@ -147,6 +154,11 @@ public function getCurrentUserId() {
 			return '';
 		}
 
+		if ($this->providerService->getSetting($provider->getId(), ProviderService::SETTING_CHECK_BEARER, '1') !== '1') {
```
juliushaertl, line 157:
```suggestion
		if ($this->providerService->getSetting($provider->getId(), ProviderService::SETTING_CHECK_BEARER, '1') !== '1') {
```

I'd vote for rather making it turned off by default, what do you think?

eneiluj: Yes right, let's make it off by default.

eneiluj:
Tests are now fixed. I don't know why it passed in #320 because there was an error there:
https://github.com/nextcloud/user_oidc/commit/5cb7be43db5479f40a18ee3d137dc0487a3b0f07#diff-45409db9d25374a7443080fb85be8ea251e80d2c2bf86c18401dd68664467ad2L109-R112

Default value for bearer check is now false.

lib/User/Backend.php
```diff
@@ -147,6 +154,11 @@ public function getCurrentUserId() {
 			return '';
 		}
 
+		if ($this->providerService->getSetting($provider->getId(), ProviderService::SETTING_CHECK_BEARER, '0') !== '1') {
+			$this->logger->error('Bearer token check is disabled for provider ' . $provider->getId());
```
juliushaertl, line 158:
```suggestion
			$this->logger->debug('Bearer token check is disabled for provider ' . $provider->getId());
```

Otherwise this would spam heavily for disabled checking

juliushaertl (commented):
Small comment on the log level otherwise LGTM :+1:

lib/User/Backend.php
juliushaertl: Amended to the existing commit.

juliushaertl (approved):

PR #326, opened 2021-09-21
Login with session token

### LINKED ISSUE #325: clients become logged out short after login
There is an closed issue for nextcloud server, where logings were invalidated within minutes: https://github.com/nextcloud/server/issues/26502

The issue was closed, as pulsejet updated their Social Login app: https://github.com/nextcloud/server/issues/26502#issuecomment-822072191

However the issue seems still present for this app ( the OIDC backend). It might not work properly with nextcloud 22.

ThomasSteinbach:
I am no PHP programmer, but this could possibly fix #325 .

I just have adapted the change from https://github.com/pulsejet/nextcloud-oidc-login/commit/3a50a43611d859daed44b065db86b16225052404

Could someone test and approve this change?

This would fix the issue, that on latest Nextcloud 22 client sessions would be invalidated after minutes. Please see #325 for more information.


lib/Controller/LoginController.php
```diff
@@ -356,8 +357,20 @@ public function code($state = '', $code = '', $scope = '') {
 		$this->logger->debug('Logging user in');
 
 		$this->userSession->setUser($user);
-		$this->userSession->completeLogin($user, ['loginName' => $user->getUID(), 'password' => '']);
+		
+		$this->userSession->getSession()->regenerateId();
+		$tokenProvider = \OC::$server->query(DefaultTokenProvider::class);
+		$this->userSession->setTokenProvider($tokenProvider);
```
juliushaertl, line 363: Is this really necessary, why would it need to use enforce the DefaultTokenProvider? 

Do you have a more detailed list of steps to reproduce the issue? Maybe you can also share your Nextcloud config in place to check for the session timeout variables, as I so far didn't manage to reproduce any issue.

lib/Controller/LoginController.php
```diff
@@ -356,8 +357,20 @@ public function code($state = '', $code = '', $scope = '') {
...
+		
 		$this->userSession->createSessionToken($this->request, $user->getUID(), $user->getUID());
+		
+		$token = $tokenProvider->getToken($this->userSession->getSession()->getId());
+
+		$this->userSession->completeLogin($user, [
+			'loginName' => $user->getUID(),
+			'password' => '',
```
juliushaertl, line 371: Might be that we should rather fix this behavior in the server or differently then as the empty password approach is also used for SAML and hasn't shown any issue there as well.

ThomasSteinbach:
@juliushaertl as I stated above, I just have adapted the change from https://github.com/pulsejet/nextcloud-oidc-login/commit/3a50a43611d859daed44b065db86b16225052404 - without exactly knowing, if it works. I am no PHP programmer and just wanted to deliver a hint of how the problem could be solved. Or was solved on a similar Nextcloud application.

PR #334, opened 2021-10-04
Remove redundant and time consuming userinfo validation

eneiluj:
As we claim the required attributes, they will be included in the token itself. If they are not, they won't be included in `userinfo` response neither.

Let's keep the related validator though.


juliushaertl (approved):
:tada:

PR #335, opened 2021-10-05
Keep userinfo validator for api calls only

eneiluj:
In some case (which I faced IRL :grin:) the uid attribute is not in the token but is returned by the userinfo endpoint.

In fact, when receiving an api call, we don't know how the OIDC token was generated and the associated claims.


lib/User/Backend.php
```diff
@@ -42,6 +43,7 @@
 class Backend extends ABackend implements IPasswordConfirmationBackend, IGetDisplayNameBackend, IApacheBackend {
 	private $tokenValidators = [
 		SelfEncodedValidator::class,
+		UserInfoValidator::class,
```
juliushaertl, line 46: Welcome back :grin: Maybe we should just make this configurable through some config.php option. We also have another use case where the custom oidc implementation could requirea custom validation class implementation.

eneiluj: Hmmm and how do we handle the case where we want to get userinfo on Bearer token validation but not on login validation?

There could be 2 config.php options, `login_userinfo_validation` and `api_userinfo_validation`. What do you think?

> We also have another use case where the custom oidc implementation could requirea custom validation class implementation.

These custom validation classes could be registered by other apps like we do for the dashboard widgets. Except `user_oidc` is not shipped by default, so it would require to emit an event in user_oidc's Application.php, right?

juliushaertl: But we only want it on the bearer token validation anyways, right? As per https://github.com/nextcloud/user_oidc/pull/334#issue-1015118675 for the login we claim the required attributes and the validator reintroduced here doesn't get called for the regular login

eneiluj: Very true.

PR #337, opened 2021-10-11
Cache provider public keys

julien-nc:
We could cache the jwks. This avoid 2 frequents requests to discovery endpoint and jwks_uri.
I don't know if the keys can change in time. I assume it's very rare. Even then, they will be refreshed when the cache expires.
Is this a good approach to cache such things?

### DIFF
--- a/lib/Service/DiscoveryService.php
+++ b/lib/Service/DiscoveryService.php
@@ -31,16 +31,22 @@ use OCP\Http\Client\IClientService;
 use Psr\Log\LoggerInterface;
 
 class DiscoveryService {
+	public const INVALIDATE_JWKS_CACHE_AFTER_SECONDS = 3600;
 
 	/** @var LoggerInterface */
 	private $logger;
 
 	/** @var IClientService */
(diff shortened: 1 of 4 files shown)


lib/Service/DiscoveryService.php
```diff
@@ -54,10 +60,21 @@ public function obtainDiscovery(Provider $provider): array {
...
+			$rawJwks = json_decode($rawJwks, true);
+			$jwks = JWK::parseKeySet($rawJwks);
+		} else {
+			$discovery = $this->obtainDiscovery($provider);
+			$client = $this->clientService->newClient();
+			$responseBody = $client->get($discovery['jwks_uri'])->getBody();
+			$result = json_decode($responseBody, true);
+			$jwks = JWK::parseKeySet($result);
+			// cache jwks
+			$this->providerService->setSetting($provider->getId(), 'jwks_cache', $responseBody);
```
juliushaertl, line 75: Would be good to also invalidate that cache when updating the provider

PR #353, opened 2021-11-29
Let .nextcloudignore skip defined paths only in root

juliushaertl:
As reported by @tsdicloud


.nextcloudignore
```diff
@@ -1,29 +1,29 @@
...
+/l10n/no-php
+/Makefile
+/&nbproject
+/node_modules
+/package.json
+/package-lock.json
+/screenshots
+/src
+/stylelint.config.js
+/tests
```
ChristophWurst, line 23: this might be the exception. there are some composer dependencies that ship tests. dropping those files might be desirable.

.nextcloudignore
```diff
@@ -1,29 +1,29 @@
...
+/.scrutinizer.yml
+/CONTRIBUTING.md
+/composer.json
+/composer.lock
+/composer.phar
+/karma.conf.js
+/krankerl.toml
+/l10n/no-php
+/Makefile
+/&nbproject
```
ChristophWurst, line 16:
```suggestion
/nbproject
```

typo?

ChristophWurst (changes requested):

.nextcloudignore
juliushaertl: Added a separate rule for `/vendor/**/tests` to be more explicit about that case

ChristophWurst (approved):

eneiluj (approved):

PR #358, opened 2021-12-20
Avoid empty session on certain redirect situations in Safari

juliushaertl:
Steps so far needed to reproduce:

- Share a file and require users to accept the share
- On Safari as the recipient
	- Click the share accept link from the mail in some webmail client (3rd party domain)
	- The link gets opened in a new tab/window 
	- The new window redirects immediately which causes the session cookie not to be stored in safari
	- Redirect to OIDC happens
	- After returning to Nextcloud, the session state is empty (likely due to https://bugs.webkit.org/show_bug.cgi?id=196375) with the cross domain redirect

Before:
User sees json response that states didn't match, because the session was empty after the redirect

After:
The user is being logged in properly

### DIFF
--- a/lib/Controller/LoginController.php
+++ b/lib/Controller/LoginController.php
@@ -134,6 +134,9 @@ class LoginController extends Controller {
 	 * @UseSession
 	 */
 	public function login(int $providerId, string $redirectUrl = null) {
+		if ($this->userSession->isLoggedIn()) {
+			return new RedirectResponse($redirectUrl);
+		}
 		$this->logger->debug('Initiating login for provider with id: ' . $providerId);
 
 		//TODO: handle exceptions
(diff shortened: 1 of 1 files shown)


lib/Controller/LoginController.php
```diff
@@ -213,6 +216,10 @@ public function login(int $providerId, string $redirectUrl = null) {
 		$url = $discovery['authorization_endpoint'] . '?' . http_build_query($data);
 		$this->logger->debug('Redirecting user to: ' . $url);
 
+		if ($this->request->isUserAgent(['/Safari/'])) {
```
juliushaertl, line 219: We can probably further only use this condition in cases when there is no redirect happening from https://github.com/nextcloud/server/blob/master/lib/private/AppFramework/Middleware/Security/SecurityMiddleware.php#L268

eneiluj:
LGTM but I don't get exactly why the bug would appear with a RedirectResponse and a not with the page which contains `<meta...`

Also, this looks strange:
```
content="0; url=' . $url . '"
```
Are you sure it's not
```
content="0" url="' . $url . '"
```
?

juliushaertl:
About the format, yes that is the right one: https://en.wikipedia.org/wiki/Meta_refresh#Examples

My assumption is that the session cookie only gets stored somewhere in the dom loading process and therefore not when only redirects. If there is one set as with this PR the cookie is properly sent with each request during the redirect chain and therefore updating the session state on the PHP side works fine.

eneiluj (approved):

tsdicloud:
I don't exactly understand why the patch does the enhanced redirect for Safari only in the `login` method and not at the end of code `method' ? (I need to upgrade our customized version which contains even more redirect cases..)

juliushaertl:
The login method is the only place where the cross-domain redirect happens without user interaction so from NC to IdP and this is the case where I managed to reproduce the issue with Safari that if the redirect is triggered from an initial page load by a 3rd-party webmail source in a new tab. Any follow up redirects that may happen are handled properly with the existing redirects through the HTTP headers.

PR #385, opened 2022-02-11
Move to IBootstrap

juliushaertl:
Avoid spamming the log and have more sane registration process


eneiluj (approved):
Tested in real conditions
* login works
* login with redirect works
* user backend is registered

Much nicer!

eneiluj:
Rebased on master

PR #411, opened 2022-04-04
Update master php testing versions

nickvergessen:
Signed-off-by: Joas Schilling <coding@schilljs.com>

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -13,7 +13,7 @@ jobs:
 
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


nickvergessen:
> Pushed a few more adjustments, the last commit we might want to still keep partly adjusted, to run at least one db backend against all php versions.

Sounds good, currently it's running 9 times and that a problem is specific to a database combined with a single php version and still not showing in unit tests sounds unlikely

juliushaertl (approved):

PR #412, opened 2022-04-05
Cache user object when checking existance

CarlSchwan:
In server:lib/private/User/Manager.php::get we always first check for
the existance of the user and then try to fetch it. Caching the user
should save us one call to the DB on each request for a OIDC user.

### DIFF
--- a/lib/Db/UserMapper.php
+++ b/lib/Db/UserMapper.php
@@ -29,14 +29,18 @@ use OCA\UserOIDC\Service\ProviderService;
 use OCP\AppFramework\Db\IMapperException;
 use OCP\AppFramework\Db\QBMapper;
 use OCP\IDBConnection;
+use OC\Cache\CappedMemoryCache;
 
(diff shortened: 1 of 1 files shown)


juliushaertl (approved):
Good catch 👍

CarlSchwan:
> Hey Carl,
> 
> I don't think the patch is really optimal as it now calls getUser twice in userExists. I would put the caching at the end of getUser like:
> 
> ```
> 	$qb->select('*')
> 			->from($this->getTableName())
> 			->where(
> 				$qb->expr()->eq('user_id', $qb->createNamedParameter($uid))
> 			);
> 
>         $dbuser = $this->findEntity($qb);
>         $this->userCache->set($uid, $dbuser);
> 	return $dbuser;
> ```

getUser in userExists didn't get called twice because the old getUser line was removed. But it indeed make even more sense to cache in getUser I created a PR for that in https://github.com/nextcloud/user_oidc/pull/435

PR #416, opened 2022-04-11
Allow non-auto provisionning

eneiluj:
We might just want to authenticate via user_oidc but let another user backend provision the user.

Auto provisioning can be toggled with a system value. For the moment it is applied to all providers (just because user_saml does it like that).
This PR brings 2 changes. For login and bearer token validation: If auto-provisioning is disabled, don't create the user and don't update its attributes with the mapped ones but still make the authentication successful.


lib/Controller/LoginController.php
```diff
@@ -329,50 +329,59 @@ public function code($state = '', $code = '', $scope = '', $error = '', $error_d
...
-			return new JSONResponse($payload);
-		}
+		$oidcSystemConfig = $this->config->getSystemValue('user_oidc', []);
+		$autoProvisionAllowed = (!isset($oidcSystemConfig['auto_provision']) || $oidcSystemConfig['auto_provision']);
+		if (!$autoProvisionAllowed) {
+			$user = $this->userManager->get($userId);
+			if ($user === null) {
+				return new JSONResponse(['Failed to provision user']);
+			}
+		} else {
```
juliushaertl, line 339: We could move the whole user provisioning (both if/else branches) to a separate provisionUser method that returns the IUser. Would make the code a bit more separated and could then return really for the auto provision to avoid one nesting level-

juliushaertl, line 339: We could move the whole user provisioning (both if/else branches) to a separate provisionUser method that returns the IUser. Would make the code a bit more separated and could then return really for the auto provision to avoid one nesting level-

juliushaertl:
Haven't tested yet, but looks sane code wise. Maybe we could add a very simple integration test case with a local user?

lib/Controller/LoginController.php
eneiluj: > and could then return really for the auto provision to avoid one nesting level-

I didn't get that part :grin:

Anyway, yep, done.

eneiluj:
Thanks to your comment about the tests I realized I could try this PR with local users (Database backend) instead of deploying an LDAP server...

It now works fine, with auto provisioning diabled, login and bearer token validation are fine and can be used to authenticate as a user who is accepted by the OIDC provider but not managed by the user_oidc user backend.

So I made a few adjustments in the user backend. It's now much simpler that what user_saml does, we don't get the list of other user backends (that wasn't working anyway) and we don't make changes to the backend's userExists method. Instead, when validating a bearer token, if auto provisioning is disabled, we check if the user exists with the userManager. Does that seem fine to you?

I'll now try to write some tests.

eneiluj:
The only way I found to make the tests pass was to surround my test by `sleep(5)`... There probably are some async stuff happening which leads to unexpected behaviors in other tests like keycloak responding 500. Any idea how to make it clean without the sleep?

So the test disables auto provisioning, creates a database user and triggers a keycloak login while making sure the mapped user id is the same as the local one. Then it checks the display name (from NC page's <head>) is the one of the local user.

lib/Controller/LoginController.php
juliushaertl: Damn autocorrection, return **early** of course ;)

juliushaertl:
> The only way I found to make the tests pass was to surround my test by sleep(5)... There probably are some async stuff happening which leads to unexpected behaviors in other tests like keycloak responding 500. Any idea how to make it clean without the sleep?

I cannot see any obvious reason, especially since the tests should not reuse any cookies with the new client+cookie jar being reset in the setUp method. Maybe the key cloak log gives some more details, but we do not dump that yet in the CI.

juliushaertl:
Pushed a commit to cleanup the test users more actively which at least when running them locally seems to work just fine without the sleeps, lets see what CI said

eneiluj:
Rebased on master.
The 500 returned by keycloak is very strange.

juliushaertl:
What is especially odd that the tests always pass locally for me.

eneiluj:
I would be happy to force merge this but then CI will always fail...

eneiluj:
I give up :grin:. Let's merge that with the sleep. Nothing I tried without the sleeps fixed the tests.

PR #437, opened 2022-05-17
Make sure any delayed avatar generation is triggered

eneiluj:
Another improvement for non-auto provisioning.

Call `$user->canChangeAvatar()` on login to make sure any delayed avatar generation is triggered.
This calls `userBackend->canChangeAvatar($uid)` which may or may not take care of avatar generation. In the `user_ldap` case, it does :grin:.


juliushaertl (approved):
Makes sense, good to know now how that gets triggered.

PR #450, opened 2022-06-29
Add admin endpoint to create user for pre-provisioning

eneiluj:
In some cases, admins might want to pre-provision users so files can be shared with them before they even logged in once.

A potential solution is to disable auto provision and use user_ldap to provision the users. Problem is then user_oidc looses the ability to update the users attributes because they are then handled by user_ldap.

This small api endpoint allows to pre-create OIDC users (and their storage) just like if they had logged in.

### DIFF
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -28,9 +28,13 @@ return [
 		['name' => 'login#login', 'url' => '/login/{providerId}', 'verb' => 'GET'],
 		['name' => 'login#code', 'url' => '/code', 'verb' => 'GET'],
 		['name' => 'login#singleLogoutService', 'url' => '/sls', 'verb' => 'GET'],
+
+		['name' => 'api#createUser', 'url' => '/user', 'verb' => 'POST'],
+
 		['name' => 'id4me#showLogin', 'url' => '/id4me', 'verb' => 'GET'],
 		['name' => 'id4me#login', 'url' => '/id4me', 'verb' => 'POST'],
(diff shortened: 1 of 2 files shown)


juliushaertl:
Different approach might be to implement user creation in the user backend so that it could make use of the provisioning api: https://github.com/nextcloud/server/blob/9ec0cb0a90ecdc7f4181ce8399bfe0c3a35842a2/lib/private/User/Manager.php#L405

lib/Controller/LoginController.php
```diff
@@ -419,6 +420,42 @@ private function provisionUser(string $userId, int $providerId, object $idTokenP
...
+		if ($email) {
+			$user->setEMailAddress($email);
+		}
+
+		if ($quota) {
+			$user->setQuota($quota);
+		}
+
+		$this->userSession->setUser($user);
+		$this->userSession->completeLogin($user, ['loginName' => $user->getUID(), 'password' => '']);
```
juliushaertl, line 454: Thinking about it again, does this have any side effect when calling this API method through an existing user session? Maybe we should explicitly setup the FS only here: 

https://github.com/nextcloud/globalsiteselector/blob/1c72d573703406a3025c2d3b9727f8ac9d774b22/lib/UserBackend.php#L123-L129

eneiluj:
New ApiController to avoid putting this in LoginController which didn't make sense.

juliushaertl (approved):

PR #451, opened 2022-06-29
Handle Ldap deleted users with non-auto provisioning

eneiluj:
If using ldap to provision users, make sure deleted users can't authenticate via login or bearer token.

### DIFF
--- a/lib/Service/LdapService.php
+++ b/lib/Service/LdapService.php
@@ -1,0 +1,75 @@
+<?php
+/*
+ * @copyright Copyright (c) 2022 Julien Veyssier <eneiluj@posteo.net>
+ *
+ * @author Julien Veyssier <eneiluj@posteo.net>
+ *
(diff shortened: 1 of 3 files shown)


lib/Service/LdapService.php
```diff
@@ -0,0 +1,55 @@
...
+	}
+
+	public function isLdapDeletedUser(IUser $user): bool {
+		$className = $user->getBackendClassName();
+		if ($className !== 'LDAP') {
+			return false;
+		}
+
+		/** @var \OCA\User_LDAP\User\DeletedUsersIndex */
+		$dui = \OC::$server->get(\OCA\User_LDAP\User\DeletedUsersIndex::class);
```
juliushaertl, line 42: Can we catch and properly handle if the class in not there?

juliushaertl (approved):

PR #464, opened 2022-07-18
Implement backchannel logout

### LINKED ISSUE #447: Implement backchannel logout
We could support backchannel logout like described in https://openid.net/specs/openid-connect-backchannel-1_0.html

The logic, as I understand it is:
* there's a way to check if it's supported by the provider (attributes of the discovery)
* on login, give the provider our session ID and a backchannel logout URL
* on logout, we call the single logout endpoint (already implemented) and the provider will call the backchannel URL with the session ID as param
* we should close the corresponding session

eneiluj:
closes #447

More commits will come with instructions to configure backchannel logout on the OP and the RP.

I tried to respect those specs as much as possible: https://openid.net/specs/openid-connect-backchannel-1_0.html

As we receive a logout token in the backchannel logout request, we get the sub and sid. IMO checking the sid is enough to identify the session.

So, on login, we store the relation between the session ID (of the ISession, not the IUserSession) and the ID token's `sid` attribute.

On backchannel logout, we get the session ID from the `sid` and we invalidate this session. Some more cleanup might be needed because the sid/sessionId links are not deleted when a session expires for example.

### DIFF
--- a/lib/Controller/LoginController.php
+++ b/lib/Controller/LoginController.php
@@ -362,9 +362,18 @@ class LoginController extends Controller {
 		$this->userSession->createRememberMeToken($user);
 
 		// for backchannel logout
-		$authToken = $this->authTokenProvider->getToken($this->session->getId());
-
-		$this->sessionMapper->createSession($idTokenPayload->sid, $idTokenPayload->sub, $idTokenPayload->iss, $authToken->getId(), $this->session->getId());
+		try {
+			$authToken = $this->authTokenProvider->getToken($this->session->getId());
+			$this->sessionMapper->createSession(
+				$idTokenPayload->sid ?? 'fallback-sid',
+				$idTokenPayload->sub ?? 'fallback-sub',
+				$idTokenPayload->iss ?? 'fallback-iss',
+				$authToken->getId(),
+				$this->session->getId()
+			);
+		} catch (InvalidTokenException $e) {
+			$this->logger->debug('Auth token not found after login');
+		}
 
 		// if the user was provisioned by user_ldap, this is required to update and/or generate the avatar
 		if ($user->canChangeAvatar()) {
--- a/lib/Controller/LoginController.php
+++ b/lib/Controller/LoginController.php
@@ -27,6 +27,8 @@ declare(strict_types=1);
 
 namespace OCA\UserOIDC\Controller;
 
+use OC\Authentication\Exceptions\InvalidTokenException;
+use OC\Authentication\Token\IProvider;
 use OCA\UserOIDC\Event\AttributeMappedEvent;
 use OCA\UserOIDC\Event\TokenObtainedEvent;
 use OCA\UserOIDC\Service\DiscoveryService;
@@ -58,6 +60,7 @@ class LoginController extends Controller {
 	private const NONCE = 'oidc.nonce';
 	public const PROVIDERID = 'oidc.providerid';
 	private const REDIRECT_AFTER_LOGIN = 'oidc.redirect';
+	private const CURRENT_SID = 'oidc.sid';
 
 	/** @var ISecureRandom */
 	private $random;
@@ -105,6 +108,10 @@ class LoginController extends Controller {
 	 * @var LdapService
 	 */
 	private $ldapService;
+	/**
+	 * @var IProvider
+	 */
+	private $authTokenProvider;
 
 	public function __construct(
 		IRequest $request,
@@ -122,6 +129,7 @@ class LoginController extends Controller {
 		ITimeFactory $timeFactory,
 		IEventDispatcher $eventDispatcher,
 		IConfig $config,
+		IProvider $authTokenProvider,
 		ILogger $logger
 	) {
 		parent::__construct(Application::APP_ID, $request);
@@ -141,6 +149,7 @@ class LoginController extends Controller {
 		$this->logger = $logger;
 		$this->config = $config;
 		$this->ldapService = $ldapService;
+		$this->authTokenProvider = $authTokenProvider;
 	}
 
 	/**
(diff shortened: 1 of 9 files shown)


lib/Controller/LoginController.php
```diff
@@ -344,6 +353,10 @@ public function code($state = '', $code = '', $scope = '', $error = '', $error_d
 		$this->userSession->createSessionToken($this->request, $user->getUID(), $user->getUID());
 		$this->userSession->createRememberMeToken($user);
 
+		// for backchannel logout
+		$this->config->setAppValue(Application::APP_ID, 'sid-' . $idTokenPayload->sid, $this->session->getId());
```
juliushaertl, line 357: Generally IConfig might have the issue that there is no proper way that we can expire this, so if users don't logout manually this will be growing forever. Maybe we should go with either caching like redis or a dedicated db table?

lib/Controller/LoginController.php
```diff
@@ -344,6 +353,10 @@ public function code($state = '', $code = '', $scope = '', $error = '', $error_d
...
+		$this->session->set(self::CURRENT_SID, $idTokenPayload->sid);
```
juliushaertl, line 358: So we actually need to store this in the session or could we just get the current authtoken somehow? Session might not be available if the remember me cookie is used, iirc.

lib/Controller/LoginController.php
eneiluj: Caching is not an option here because we need to be sure the value is stored and caching might not have been configured in the instance.

I thought I could find a trick to cleanup app config values but you're right, it might not be possible.

Let's go for a dedicated table then. The cleanup mechanism would then be a cron job deleting sid/sessionId links that are older than...one day? What do you think?

eneiluj: Or a smarter way to clean things up is to delete `oc_user_oidc_sessions` (the future table) entries referring to entries that do not exists anymore in `oc_authtoken`.

eneiluj:
There's now a `oc_user_oidc_sessions` table storing all we need for backchannel logout:
* sid, sub and iss from the login OIDC ID token
* the authtoken ID to get the token object back and invalidate the NC session
* the NC session ID (only used for cleanup purposes on classic logout)
* the oidc session creation time

About the cleanup:

> Should probably depend on the values of remember_login_cookie_lifetime / session_lifetime in config.php.
> Or do you see a reason to make that configurable on its own depending on possible lifetime on the IdP?

I don't know. I think we are not doing things properly in the first place :grin:. Independently from the backchannel logout, the NC session should expire when the OIDC ID token expires (or before, if session_lifetime system config says so). Then as we would have the same expiration dates for the OIDC token and the NC auth token, we could store this expiration date in `oc_user_oidc_sessions` and use it in a cron job to clean things up. What do you think?

eneiluj:
@juliushaertl Cleanup is there.

lib/Cron/CleanupSessions.php
```diff
@@ -0,0 +1,64 @@
...
+	 */
+	private $config;
+
+	public function __construct(ITimeFactory  $time,
+								IConfig $config,
+								SessionMapper $sessionMapper) {
+		parent::__construct($time);
+		$this->sessionMapper = $sessionMapper;
+		// daily
+		$this->setInterval(24 * 60 * 60);
```
juliushaertl, line 49: We can mark this time insensitive https://docs.nextcloud.com/server/latest/developer_manual/basics/backgroundjobs.html#heavy-load-and-time-insensitive if it is only running once a day

lib/Migration/Version01021Date20220719114355.php
```diff
@@ -0,0 +1,66 @@
...
+			'notnull' => true,
+			'length' => 200,
+		]);
+		$table->addColumn('created_at', Types::BIGINT, [
+			'notnull' => true,
+			'length' => 20,
+			'unsigned' => true,
+		]);
+		$table->setPrimaryKey(['id']);
+		$table->addUniqueIndex(['sid']);
```
juliushaertl, line 62: We should add an index for created_at as the background job filters by that

lib/Db/SessionMapper.php
```diff
@@ -0,0 +1,151 @@
...
+	 * @param string $ncSessionId
+	 * @return int
+	 * @throws \OCP\DB\Exception
+	 */
+	public function deleteFromNcSessionId(string $ncSessionId): int {
+		$qb = $this->db->getQueryBuilder();
+
+		$qb->delete($this->getTableName())
+			->where(
+				$qb->expr()->eq('nc_session_id', $qb->createNamedParameter($ncSessionId, IQueryBuilder::PARAM_STR))
```
juliushaertl, line 89: nc_session_id should also get an index 😉

lib/Db/SessionMapper.php
```diff
@@ -0,0 +1,151 @@
...
+			->where(
+				$qb->expr()->lt('created_at', $qb->createNamedParameter($minCreationTimestamp, IQueryBuilder::PARAM_INT))
+			);
+		$qb->executeStatement();
+	}
+
+	/**
+	 * @return Session[]
+	 */
+	public function getSessions() {
```
juliushaertl, line 111: Seems unused

juliushaertl (changes requested):
A few more comments, but looks good in general :)

eneiluj:
Thanks for the review! Agreed, done.

lib/Cron/CleanupSessions.php
```diff
@@ -45,9 +46,10 @@ public function __construct(ITimeFactory  $time,
 								SessionMapper $sessionMapper) {
 		parent::__construct($time);
 		$this->sessionMapper = $sessionMapper;
+		$this->config = $config;
 		// daily
 		$this->setInterval(24 * 60 * 60);
-		$this->config = $config;
+		$this->setTimeSensitivity(IJob::TIME_INSENSITIVE);
```
juliushaertl, line 52: May need a method_exists check if we want to keep the app compatible below Nextcloud 24.

eneiluj: The method is available in server's stable23 and stable22 (even if @since says different :grin:). So it would just be for NC 21. We can also drop support for NC 21. What do you think?

juliushaertl: Oh right, we back ported it, but may still throw then on older patch releases

eneiluj: Yep

juliushaertl (approved):

Issue #449, opened 2022-06-29
Can not logout via Keycloak - Missing parameters id_token_hint


Lycsona:
I have an issue with logout via Keycloak.

When a user clicks on the logout button, their see:

`Missing parameters: id_token_hint`

The Keycloak logout URL is used properly, but **id_token_hint** which is required is not set. 
This token is issued after the user signs in as an **id_token** value from Keycloak response of auth endpoint. 

Could it be fixed, please? Or could you help to clarify it? Because I am not able to use logout.

Igortorrente:
Same issue

DaphneMuller:
@julien-nc can you look into an effort estimate for this one? It's the top-voted issue among all our team's apps

PR #373, opened 2022-01-20
Implement single logout

eneiluj:
Here is an attempt to implement single logout using discovery's end_session_endpoint.
It needs more error handling and some discussions about `Backend::isSessionActive()`.

Issue #336, opened 2021-10-11
Logout with endsession endpoint


antoniogfs:
Would it be possible to implement the logout process using the endsession endpoint as described by the OIDC standard?
Currently this is not supported.

Mr-Markus:
There is a PR https://github.com/nextcloud/user_oidc/pull/373 from @eneiluj 
Are you still working on it and how is the status?

eneiluj:
@Mr-Markus Yes it's still a WIP. It's coming soon though as I feel the implementation is mostly done. I just have some doubts I want to clarify.

joshtrichards:
Added in #373

PR #493, opened 2022-08-26
Add client_id and id_token_hint to IdP logout

ubipo:
_Edited to add links to docs/code examples_

This adds two parameters to the end_session_endpoint IdP URL which the
user gets redirected to when singleLogout is triggered.

These paramters are:
- client_id: the client ID of the current session's provider. 'OPTIONAL'
  as per the relevant OpenID specification.
- id_token_hint: the raw id_token that was obtained during the code
  callback of this session's login flow (set in session variable `oidc.id_token`). 'RECOMMENDED' by the relevant OpenID specification [1].

Some providers (e.g. node-oidc-provider [2] and Keycloak [3]) require this when using the code OAuth flow.

Because passing id_token_hint reveals the id_token to the user agent, a
app setting was also added to optionally turn this behaviour off (default is
turned on).

Builds upon PR https://github.com/nextcloud/user_oidc/pull/373 / issue https://github.com/nextcloud/user_oidc/issues/336
Fixes issue https://github.com/nextcloud/user_oidc/issues/449

[1]: https://openid.net/specs/openid-connect-rpinitiated-1_0.html#RPLogout
[2]: https://github.com/panva/node-oidc-provider/blob/c243bf6b6663c41ff3e75c09b95fb978eba87381/lib/actions/end_session.js#L32
[3]: https://www.keycloak.org/docs/latest/release_notes/index.html#oidc-logout-changes

Signed-off-by: Pieter Fiers <pieter@pfiers.net>

### DIFF
--- a/lib/Service/ProviderService.php
+++ b/lib/Service/ProviderService.php
@@ -34,6 +34,7 @@ use OCP\IConfig;
 
 class ProviderService {
 	public const SETTING_CHECK_BEARER = 'checkBearer';
+	public const SETTING_SEND_ID_TOKEN_HINT = 'sendIdTokenHint';
 	public const SETTING_UNIQUE_UID = 'uniqueUid';
 	public const SETTING_MAPPING_UID = 'mappingUid';
 	public const SETTING_MAPPING_UID_DEFAULT = 'sub';
@@ -43,6 +44,11 @@ class ProviderService {
 	public const SETTING_EXTRA_CLAIMS = 'extraClaims';
 	public const SETTING_JWKS_CACHE = 'jwksCache';
 	public const SETTING_JWKS_CACHE_TIMESTAMP = 'jwksCacheTimestamp';
+	private const BOOLEAN_SETTINGS = array(
+		self::SETTING_UNIQUE_UID,
+		self::SETTING_CHECK_BEARER,
+		self::SETTING_SEND_ID_TOKEN_HINT
+	);
 
 	/** @var IConfig */
 	private $config;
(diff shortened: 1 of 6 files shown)


lib/Service/ProviderService.php
```diff
@@ -43,6 +44,10 @@ class ProviderService {
 	public const SETTING_EXTRA_CLAIMS = 'extraClaims';
 	public const SETTING_JWKS_CACHE = 'jwksCache';
 	public const SETTING_JWKS_CACHE_TIMESTAMP = 'jwksCacheTimestamp';
+	private const BOOLEAN_SETTINGS = array(
+		self::SETTING_UNIQUE_UID, self::SETTING_CHECK_BEARER,
+		self::SETTING_SEND_ID_TOKEN_HINT
+	);
```
eneiluj, line 50:
```suggestion
	private const BOOLEAN_SETTINGS = [
		self::SETTING_UNIQUE_UID,
		self::SETTING_CHECK_BEARER,
		self::SETTING_SEND_ID_TOKEN_HINT,
	];
```

src/components/SettingsForm.vue
```diff
@@ -108,6 +108,12 @@
 		<p class="settings-hint">
 			{{ t('user_oidc', 'Do you want to allow API calls and WebDav request that are authenticated with an OIDC ID token or access token?') }}
 		</p>
+		<CheckboxRadioSwitch :checked.sync="localProvider.settings.sendIdTokenHint" wrapper-element="div">
+			{{ t('user_oidc', 'Send ID token hint on logout') }}
+		</CheckboxRadioSwitch>
+		<p class="settings-hint">
+			{{ t('user_oidc', 'Should the ID token be included as a parameter (id_token_hint) of the provider\'s logout URL? The user will be redirected to this URL after logout in Nextcloud. Enabling this exposes the ID token to the user agent, which may not be necessary depending on the particular OIDC provider used.') }}
```
eneiluj, line 115:
```suggestion
			{{ t('user_oidc', 'Should the ID token be included as a the id_token_hint GET parameter in the OpenID logout URL? User are redirected to this URL after logging out of Nextcloud. Enabling this setting exposes the OIDC ID token to the user agent, which may not be necessary depending on the OIDC provider.') }}
```

eneiluj (changes requested):
Nice! Thanks for the PR. Nicely done.

I had trouble finding doc about why the client_id is needed but here it is: https://www.keycloak.org/docs/latest/release_notes/index.html#oidc-logout-changes

Could you sign your commits and rebase your branch on this project's master?

ubipo:
Thanks for the review! I've implemented the suggestions, rebased onto master, signed the commit, and added some more info on the use of the parameters. Should be good to go :)

eneiluj (approved):
:+1:

PR #495, opened 2022-08-31
Enforce https

eneiluj:
Performing the login flow with HTTP reveals/exposes secret information. Login and code endpoints are now rejecting HTTP requests.

[OpenId specs regarding TLS](https://openid.net/specs/openid-connect-core-1_0.html#TLSRequirements) are not very clear but can be interpreted as "better use HTTPS".

I took the opportunity to make some polishing to the Login controller. I hope this is not too out of scope :grin:.


juliushaertl (approved):
Looks good, now running integration tests in debug mode should also pass again :)

PR #498, opened 2022-09-06
Fix mistake in user backend

eneiluj:
Produces a crash on bearer token validation.


mejo- (approved):
Looks good codewise. I didn't test though.

Issue #475, opened 2022-08-08
Excess "Could not find unique token validation" message logging


martin-nohava:
**Overview**
- One OpenID provider is configured
- WebDAV, CalDAC, clients are connected via App Tokens

**Problem**
When WebDAV/CalDAC request is made `user_oicd` logs "Could not find unique token validation" message instead of ignoring it.
On average, several such messages are logged per minute and they make the `nextcloud.log` file very large.

schewara:
Same behavior can be observed in my Testing Environment as well.

benediktkr:
seems to happen here: https://github.com/nextcloud/user_oidc/blob/master/lib/User/Backend.php#L299

goes through the oidc login providers, i think checking if the token is issued from any of them. 

![image](https://user-images.githubusercontent.com/174994/190384350-905f48ff-8d10-4315-9daf-aec0884b7e4c.png)

i have the "Check bearer token on API and WebDav requests" option un-checked, but it still seems to be trying to do that?

joshtrichards:
Fixed in #500 / #519

PR #500, opened 2022-09-15
change log message to debug to address over-logging in #475

benediktkr:
see issue #475, this happens when "Check bearer token on API and WebDav requests" is disabled and writes to the log for every webdav/caldav request (which can be a lot), seems like this should rather be logged as `debug`

### DIFF
--- a/lib/User/Backend.php
+++ b/lib/User/Backend.php
@@ -296,7 +296,7 @@ class Backend extends ABackend implements IPasswordConfirmationBackend, IGetDisp
 			}
 		}
 
-		$this->logger->error('Could not find unique token validation');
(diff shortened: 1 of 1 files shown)


julien-nc (approved):
:+1: Thanks!

julien-nc:
@benediktkr Could you sign your commit?

julien-nc:
If you can't, just tell me and i'll recommit a signed one.

julien-nc:
Closed in favor of #519

benediktkr:
@julien-nc apologies, missed your comment. i could have committed a signed commit if i had seen it in time, but i see that its been taken care of now anyway :)

PR #502, opened 2022-09-20
Extends provisioning of users

MarvinOehlerkingCap:
This PR adds functionality for auto provision user when accessing API and WebDav with Bearer token.

New settings for provider:

- groupProvisioning (This will create and update the users groups depending on the groups claim in the id token.)
- mappingGroups (Claim in the idToken which will be used for provisiong of groups.)
- providerBasedId (Use provider identifier as prefix for ids.)
- bearerProvisioning (This automatically provisions the user, when sending API and WebDav Requests with a Bearer token. Auto provisioning and Bearer token check have to be activated for this to work.)

Some code improvements:
- Moved provisioning code from Controller to a separate service


kronn:
This looks like everything I want from this plugin. Especially the ability to pass groups and have them updated later is high on my wishlist. Is there anything I can do to help this along?

kronn (approved):
Looks good to me, I like especially the group-updating.

lib/User/Backend.php
```diff
@@ -252,45 +252,52 @@ public function getCurrentUserId() {
 				// find user id through different token validation methods
 				foreach ($this->tokenValidators as $validatorClass) {
 					$validator = \OC::$server->get($validatorClass);
-					$userId = $validator->isValidBearerToken($provider, $headerToken);
-					if ($userId) {
+					$sub = $validator->isValidBearerToken($provider, $headerToken);
+					if ($sub) {
```
julien-nc, line 256: I agree `$userId` was not appropriated here but `$sub` is confusing as it's an attribute of the ID token but we might use/map another one.

How about `$oidcUserId` or `$tokenUserId`?

Same remark for all appearances of `$sub` :grin:.

src/components/SettingsForm.vue
```diff
@@ -100,18 +100,43 @@
...
+		</p>
 		<CheckboxRadioSwitch :checked.sync="localProvider.settings.checkBearer" wrapper-element="div">
 			{{ t('user_oidc', 'Check Bearer token on API and WebDav requests') }}
 		</CheckboxRadioSwitch>
 		<p class="settings-hint">
 			{{ t('user_oidc', 'Do you want to allow API calls and WebDav request that are authenticated with an OIDC ID token or access token?') }}
 		</p>
+		<CheckboxRadioSwitch :checked.sync="localProvider.settings.bearerProvisioning" wrapper-element="div">
+			{{ t('user_oidc', 'Auto provision user when accessing API and WebDav with Bearer token') }}
+		</CheckboxRadioSwitch>
```
julien-nc, line 136: This Checkbox's state could be disabled if "Check bearer token" is not checked.

src/components/SettingsForm.vue
```diff
@@ -100,18 +100,43 @@
 				type="text"
 				placeholder="quota">
 		</p>
+		<p>
+			<label for="mapping-quota">{{ t('user_oidc', 'Groups mapping') }}</label>
+			<input id="mapping-quota"
+				v-model="localProvider.settings.mappingGroups"
+				type="text"
+				placeholder="groups">
```
julien-nc, line 108: This input's state could be disabled if "Use group provisioning" is not checked.

julien-nc (changes requested):
Wow this is awesome.

A few remarks though.

I'll try this in a real setup soon.

julien-nc:
Oh and could you rebase on master?

lib/Service/IdService.php
```diff
@@ -0,0 +1,40 @@
+<?php
+
+namespace OCA\UserOIDC\Service;
+
+use OCA\UserOIDC\Db\ProviderMapper;
+
+class IdService {
```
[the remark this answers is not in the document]
MarvinOehlerkingCap: The id service is also used to generate ids for the groups. So `LocalIdService`?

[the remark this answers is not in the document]
julien-nc: :+1:

lib/Service/ProvisioningService.php
```diff
@@ -0,0 +1,174 @@
...
+		$event = new AttributeMappedEvent(ProviderService::SETTING_MAPPING_GROUPS, $idTokenPayload, json_encode($groupsData));
+		$this->eventDispatcher->dispatchTyped($event);
+		$this->logger->debug('Group mapping event dispatched');
+
+		if ($event->hasValue()) {
+			$groups = json_decode($event->getValue());
+			$userGroups = $this->groupManager->getUserGroups($user);
+			$syncGroups = [];
+
+			foreach ($groups as $k => $v) {
```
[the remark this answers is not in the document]
MarvinOehlerkingCap: groups are not directly specified by the openid standard. this implementation supports two different kinds of formatted arrays. one of them is a general list of names and the other is a more nextcloud specific one, which supports setting a separate group id and display name. there is also the option to listen for the `AttributeMappedEvent` and map your custom format to one of these formats.

I added a small example of the supported formats to the group provisioning field on the admin settings page.

lib/Service/ProvisioningService.php
```diff
@@ -0,0 +1,174 @@
...
+	public function provisionUserGroups(IUser $user, int $providerId, object $idTokenPayload): void {
+		$groupsAttribute = $this->providerService->getSetting($providerId, ProviderService::SETTING_MAPPING_GROUPS, 'groups');
+		$groupsData = $idTokenPayload->{$groupsAttribute} ?? null;
+
+		$event = new AttributeMappedEvent(ProviderService::SETTING_MAPPING_GROUPS, $idTokenPayload, json_encode($groupsData));
+		$this->eventDispatcher->dispatchTyped($event);
+		$this->logger->debug('Group mapping event dispatched');
+
+		if ($event->hasValue()) {
+			$groups = json_decode($event->getValue());
```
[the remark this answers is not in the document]
MarvinOehlerkingCap: the groupsData, like any other extracted field from the token, is dispatched to an event that may modify and map the data. the event only allows setting and returning a value as string so it has to be encoded somehow. we chose json since openid tokens uses jwt and it is easy to convert.

lib/User/Backend.php
MarvinOehlerkingCap: I renamed it to `tokenUserId`

lib/Service/ProvisioningService.php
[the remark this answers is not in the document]
julien-nc: :+1:

lib/Service/ProvisioningService.php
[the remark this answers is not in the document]
julien-nc: :+1:

julien-nc:
@MarvinOehlerkingCap Thanks for the changes. Missing things:
* [ ] @julien-nc to try it in a real environment
* [ ] rebase on master

lib/Service/ProvisioningService.php
```diff
@@ -0,0 +1,176 @@
...
+			$userGroups = $this->groupManager->getUserGroups($user);
+			$syncGroups = [];
+
+			foreach ($groups as $k => $v) {
```
julien-nc, line 138: With group provisioning enabled but when the IdP does not return the desired attributes, the event data is set to null, then we check if `$event->hasValue()` which is true and then we pass null to `foreach`.

It looks like something that could be avoided :grin:.

julien-nc (changes requested):
We're getting there. Apart from the inline comment, it looks good to me.

Could you rebase on master?
You will have to adjust to the `@nextcloud/vue` bump with the new component names.
I can do the rebase if you prefer.

julien-nc:
Side note: I got some 500 responses from Keycloak when setting user attributes of type JSON that don't have the object keys between parenthesis. Not our problem it but might be tricky to debug for admins.

lib/Service/ProvisioningService.php
MarvinOehlerkingCap: I added a simple null check for `$event->getValue()`

src/components/SettingsForm.vue
```diff
@@ -100,18 +100,44 @@
...
 			{{ t('user_oidc', 'By default every user will get a unique userid that is a hashed value of the provider and user id. This can be turned off but uniqueness of users accross multiple user backends and providers is no longer preserved then.') }}
 		</p>
+		<NcCheckboxRadioSwitch :checked.sync="localProvider.settings.providerBasedId" wrapper-element="div">
+			{{ t('user_oidc', 'Use provider identifier as prefix for ids') }}
+		</NcCheckboxRadioSwitch>
+		<p class="settings-hint">
+			{{ t('user_oidc', 'To keep ids in plain text, but also preserve uniqueness of them across multiple providers, a prefix with the providers name is added.') }}
+		</p>
+		<NcCheckboxRadioSwitch :checked.sync="localProvider.settings.groupProvisioning" wrapper-element="div">
+			{{ t('user_oidc', 'Use group provisioning. Format of the groups claim value should be [{gid: "1", displayName: "group1"}, ...] or ["group1", "group2", ...]') }}
```
julien-nc, line 124: Could you move "Format of the groups claim value should be..." to the settings hint?

julien-nc (changes requested):
Thanks for the rebase and adjustments.

One more small change :grin:.

Also, @juliushaertl, could you make a quick additional review?

Loki-Afro:
@julien-nc any update on this?

yscialom:
Yet another user excited by this PR.
Any update? Available for tests in real-life environment if need be.

julien-nc (approved):

kronn:
Where would the integration docs be located? I managed to integrate our software with Nextcloud using this plugin and would like to add my notes there. Basically: you need to pass the data for the provisioned user as part of the `id_token`.

kronn:
I used the current main branch and merged this PR. I was able to build a new version with `krankerl` and could run it without problems. I could integrate with my oauth-provider which can talk OIDC and I could pass groups along. I was happy. 

Now I have a hosted nextcloud in which I cannot easily installed custom apps. I may get shell-access or some help with that, but I would prefer having a released version with this PR merged. I also want the other fixes that went into the main branch. 

Is it possible to do a merge and a release of this? How can I help to see this done?

alweber-cap:
pinging @juliushaertl for approve/merge

juliushaertl (approved):

juliushaertl:
Tested and works. Thanks a lot for this awesome contribution.

kronn:
Thank you for the merge, this means a lot to me, @juliushaertl.

I saw that the actions on main all passed successfully. :tada: 
How is the process for releasing a new version to the app store? Should I open a github issue for it?

juliushaertl:
I'll check this with @julien-nc on if there are any other pending topics, but then I'd see no blocker for a release :)

Zegorax:
Hello everyone, thank you very much for this feature.

I implemented it on my side but I get the following error in the logs:

```
Error: foreach() argument must be of type array|object, null given at /var/www/html/apps/user_oidc/lib/Service/ProvisioningService.php#146
```

Zegorax:
> Hello everyone, thank you very much for this feature.
> 
> I implemented it on my side but I get the following error in the logs:
> 
> ```
> Error: foreach() argument must be of type array|object, null given at /var/www/html/apps/user_oidc/lib/Service/ProvisioningService.php#146
> ```

For anyone else having this problem, check that you have the correct OIDC scopes requested. In my case it solved the problem.

yscialom:
Hello, I install `user_oidc` via `occ` (script included in my nextcloud docker image):
```
occ user_oidc:provider Authelia \
    --clientid="(redacted)" \
    --clientsecret="(redacted)" \
    --discoveryuri="https://auth.(redacted)/.well-known/openid-configuration" \
    --scope="openid groups email profile" \
    --unique-uid=0 \
    --mapping-uid="preferred_username" \
    --no-interaction
```

I cannot find, neither in the doc, neither in `occ user_oidc:provider --help` neither in this PR's code how to "check" the "use group provisioning" option from `occ`.

Could you help @juliushaertl?

arnegns:
@yscialom You can set it via nextcloud config like this:

  "user_oidc": {
                    "allow_multiple_user_backends": "0",
                    "userinfo_bearer_validation": "false",
                    "selfencoded_bearer_validation": "true",
                    "types": "authentication",
                    "enabled": "yes",
                    "provider-1-mappingQuota": "",
                    "provider-1-extraClaims": "",
                    "provider-1-jwksCache": "",
                    "provider-1-jwksCacheTimestamp": "",
                    "provider-1-uniqueUid": "0",
                    "provider-1-checkBearer": "1",
                    "provider-1-bearerProvisioning": "1",
                    "provider-1-mappingUid": "sub",
                    "provider-1-mappingDisplayName": "name",
                    "provider-1-mappingEmail": "email",
                    "provider-1-mappingGroups": "groups",
                    "provider-1-**groupProvisioning**": "1",
                    "provider-1-providerBasedId": "1",
                    "id4me_enabled": "0"
                 },

yscialom:
@arnegns 
I managed, indeed, to set it through `occ`:
```
occ config:app:set user_oidc provider-1-groupProvisioning --value 1
```

I find it regrettable though to have to "guess" or "find" what my newly created provider name (here `provider-1-groupProvisioning`) is for `occ config`. It makes it harder that it could be.

MohammedNoureldin:
Hi, could anybody explain what we should change (the most basic changes) in the IdP (let us say Keycloak) to get the groups in NC assigned automatically as in the IdP?

chamabreu:
AFAIK real mapping to existinging (eg admin) nextcloud groups is not possible for now.

I have this problem, too, that i need an extra admin account to manage my nextcloud, which is not provided via keycloak.

MohammedNoureldin:
Thank you for you reply, @chamabreu.

I guess if we write a simple custom protocol mapper, which converts Administrator group to admin in the groups claim, can achieve what we want. Doesn't it?

The other option is simply implementing a field called admin IdP group(s), which is used to tell which group of IdP can be handled as admin.

PR #511, opened 2022-10-14
Bump @nextcloud/vue, vue and vue-template-compiler

julien-nc:
The ones from dependabot are outdated.

A few necessary adjustments in the provider list item:
* action buttons were streched, they are now aligned to the top
* hover color was the same for the provider item and the action buttons. For those buttons, `--color-background-hover` is now overridden by the value of `--color-background-darker`

### DIFF
--- a/.github/workflows/phpunit.yml
+++ b/.github/workflows/phpunit.yml
@@ -39,7 +39,7 @@ jobs:
 
     services:
       postgres:
-        image: postgres
+        image: postgres:14
         ports:
           - 4445:5432/tcp
         env:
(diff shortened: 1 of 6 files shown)


juliushaertl (approved):

.github/workflows/phpunit.yml
```diff
@@ -39,7 +39,7 @@ jobs:
 
     services:
       postgres:
-        image: postgres
+        image: postgres:14
```
juliushaertl, line 42: Mind to also adjust in the integration tests right away? :)

julien-nc: :+1:

PR #518, opened 2022-10-21
Fix generated authorization url

julien-nc:
Fix generated authorization url when the one from the discovery contains GET params: we merge them with the ones we add.
We keep sanitizing everything before the GET params.

@lauritzh I'm interested in your opinion on that :grin:.

### DIFF
--- a/lib/Controller/LoginController.php
+++ b/lib/Controller/LoginController.php
@@ -290,9 +290,7 @@ class LoginController extends Controller {
 			return $response;
 		}
 
-		// sanitize the authorization_endpoint
-		$authorizationEndpoint = htmlentities(filter_var($discovery['authorization_endpoint'], FILTER_SANITIZE_URL), ENT_QUOTES);
-		$authorizationUrl = $authorizationEndpoint . '?' . http_build_query($data);
+		$authorizationUrl = $this->buildAuthorizationUrl($discovery['authorization_endpoint'], $data);
 
 		$this->logger->debug('Redirecting user to: ' . $authorizationUrl);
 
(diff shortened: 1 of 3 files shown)


lib/Controller/LoginController.php
```diff
@@ -305,6 +303,37 @@ public function login(int $providerId, string $redirectUrl = null) {
...
+			. (isset($parsedUrl['port']) ? ':' . $parsedUrl['port'] : '')
+			. ($parsedUrl['path'] ?? '');
+
+		$queryParams = $extraGetParameters;
+		if (isset($parsedUrl['query'])) {
+			$params = explode('&', $parsedUrl['query']);
+			foreach ($params as $p) {
+				$keyValue = explode('=', $p);
+				if (count($keyValue) === 2) {
+					$queryParams[$keyValue[0]] = $keyValue[1];
```
juliushaertl, line 326: Should we still call htmlentities on the key and value?

lib/Controller/LoginController.php
```diff
@@ -305,6 +303,37 @@ public function login(int $providerId, string $redirectUrl = null) {
...
+		$urlWithoutParams =
+			(isset($parsedUrl['scheme']) ? $parsedUrl['scheme'] . '://' : '')
+			. ($parsedUrl['host'] ?? '')
+			. (isset($parsedUrl['port']) ? ':' . $parsedUrl['port'] : '')
+			. ($parsedUrl['path'] ?? '');
+
+		$queryParams = $extraGetParameters;
+		if (isset($parsedUrl['query'])) {
+			$params = explode('&', $parsedUrl['query']);
+			foreach ($params as $p) {
```
juliushaertl, line 323: Could also use https://www.php.net/manual/en/function.parse-str.php to clean up the code a bit :)

lib/Controller/LoginController.php
lauritzh: https://github.com/nextcloud/user_oidc/pull/518/files#diff-35a02217ab68218fc7a45f4a7b5483d63a0ea6b80973c2938f73f181845178a0R333

As the final URL is encoded, probably not needed?

I am still unsure whether it would not be better to move the `htmlentities()` to the following line where the HTML is crafted, though. 🙈

lauritzh:
The fix looks good to me and should fully remediate the reported XSS. 👍 
My comment above has no security implications :) 

Have a great weekend and best regards,
Lauritz

lib/Controller/LoginController.php
julien-nc: @juliushaertl Like Php doc says:

> http_build_query — Generate URL-encoded query string

So it seems enough to me.

@lauritzh 

> I am still unsure whether it would not be better to move the htmlentities() to the following line where the HTML is crafted, though.

We can't because, for example, it will replace `&` by `&amp;` in the query string.

lauritzh: Ahh I see. Thank you very much for clarification :)

lib/Controller/LoginController.php
julien-nc: :+1: Why making things simple when we can make them complex? :sweat_smile: 

The `buildAuthorizationUrl` is now in the `DiscoveryService`. There are some basic tests covering this method.

juliushaertl (approved):

PR #535, opened 2022-11-17
Sync LDAP user with non-auto provisioning

julien-nc:
With auto provisioning disabled, on login, we want the LDAP user list to be up to date. We already call `$this->userManager->search($userId)` but this triggers an LDAP query with a wildcard. As some LDAP attributes don't support wildcards, we need a better/safer way to make sure this specific user is synced to then be able to check if it exists locally with `$this->userManager->get($userId)`.

`OCA\User_LDAP\User_Proxy\loginName2UserName` is now called.

I still need to deploy a test environment to check this change is really effective.

### DIFF
--- a/lib/Controller/LoginController.php
+++ b/lib/Controller/LoginController.php
@@ -461,6 +461,7 @@ class LoginController extends Controller {
 			// in case user is provisioned by user_ldap, userManager->search() triggers an ldap search which syncs the results
 			// so new users will be directly available even if they were not synced before this login attempt
 			$this->userManager->search($userId);
+			$this->ldapService->syncUser($userId);
 			// when auto provision is disabled, we assume the user has been created by another user backend (or manually)
 			$user = $this->userManager->get($userId);
 			if ($this->ldapService->isLdapDeletedUser($user)) {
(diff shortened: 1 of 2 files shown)


noumar:
Hi @julien-nc and @juliushaertl,

Sorry for jumping in on this PR.

I wanted to comment that I have implemented similar functionality in [this commit](https://github.com/noumar/user_oidc/commit/ac00cee9f9fc05d7804ec86326525a6ec4f592bb), but it does more since it allows to use the returned "mapped uid".

Would you be willing to consider my solution instead?
If so, I can make a new PR. Which I've been meaning to do.

julien-nc:
> Sorry for jumping in on this PR.

@noumar Well on the contrary, thanks for jumping in.

> Would you be willing to consider my solution instead?

I don't know, in your PR you consider LDAP is used for sure for the provisioning which might not be the case in all setups.
A problematic use case would be:
The LDAP mapping (using the OIDC user ID) is successful and returns an LDAP user ID. We then get the user from the user manager from the LDAP user ID. What if another user backend could provision a user from the OIDC user ID?
Also, what if the LDAP mapping fails?

Maybe we could optionally use your logic if we know for sure users are provisioned with LDAP (a system config for example).
My opinion is: Let's deal with the sync issue first, then you could make another PR going in your direction and we could discuss there. What do you think?

noumar:
Sure. Sounds good.

julien-nc:
This has now been successfully tested on 2 different environments.
@come-nc Do you confirm these changes reflect our discussion?

PR #539, opened 2022-11-25
Switch to debug logging for bearer token checks

juliushaertl:
- Only log missing bearer token at debug level
- Log missing providers as debug

Otherwise this may produce unexpected errors in the log even if nothing is really wrong


julien-nc (approved):
:see_no_evil:

PR #541, opened 2022-12-02
Fix DB indexes

julien-nc:
Index on user_oidc_sessions.created_at which not be unique.

Changes:
* New migration step fixing all indexes if needed
* Changes in previous migration steps (only effective for app fresh install) to keep it easy to read
* Bump the app version number to trigger the new migration step


lib/Migration/Version01022Date20221202161257.php
```diff
@@ -0,0 +1,63 @@
...
+
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		if ($schema->hasTable('user_oidc_sessions')) {
+			$table = $schema->getTable('user_oidc_sessions');
+			$indexes = $table->getIndexes();
+			foreach ($indexes as $index) {
+				// fix created_at index which is not unique
+				if ($index->isUnique() && $index->hasColumnAtPosition('created_at')) {
```
juliushaertl, line 36: Shouldn't that also check for the name of the index? to not remove and add the one fixed in the previous migration on fresh installations? ;)

lib/Migration/Version01022Date20221202161257.php
```diff
@@ -0,0 +1,63 @@
...
+			$indexes = $table->getIndexes();
+			foreach ($indexes as $index) {
+				// rename index on identifier if needed
+				if ($index->isUnique() && $index->hasColumnAtPosition('identifier') && $index->getName() !== 'user_oidc_prov_idtf') {
+					$table->renameIndex($index->getName(), 'user_oidc_prov_idtf');
+				}
+			}
+		}
+
+		return $schema;
```
juliushaertl, line 61: Maybe you can have a variable that gets set when any index operation is done, so it could return null if all indexes are ok.

lib/Migration/Version01022Date20221202161257.php
julien-nc: As it check if it's unique, no risk to remove the correct one potentially added with a fresh install.

lib/Migration/Version01022Date20221202161257.php
julien-nc: In the config app settings?

I think it's nice to explicitly check everything instead of having a flag that can be outdated if we (I :grin:) do more mistakes.

juliushaertl: No in the code here, so that $schema is only returned if any of the above if conditions actually triggers a schema change.

lib/Migration/Version01022Date20221202161257.php
juliushaertl: Right :)

lib/Migration/Version01022Date20221202161257.php
julien-nc: Oh ok. Done.

juliushaertl (approved):
👍 Just needs another round for fixing the code style ;)

PR #557, opened 2023-01-09
Add system config flag to deal with external ssl termination

julien-nc:
If the flag is true, allow login via http and pass a https redirect_uri to the provider on login.

### DIFF
--- a/lib/Controller/LoginController.php
+++ b/lib/Controller/LoginController.php
@@ -205,7 +205,9 @@ class LoginController extends Controller {
 		if ($this->userSession->isLoggedIn()) {
 			return new RedirectResponse($redirectUrl);
 		}
-		if (!$this->isSecure()) {
+		$appSystemConfig = $this->config->getSystemValue('user_oidc', []);
+		$externalSslTermination = $appSystemConfig['external_ssl_termination'] ?? false;
+		if (!$externalSslTermination && !$this->isSecure()) {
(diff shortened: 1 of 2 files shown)


lib/Controller/LoginController.php
```diff
@@ -260,11 +262,15 @@ public function login(int $providerId, string $redirectUrl = null) {
 			}
 		}
 
+		$redirectUrl = $this->urlGenerator->linkToRouteAbsolute(Application::APP_ID . '.login.code');
+		if ($externalSslTermination) {
+			$redirectUrl = preg_replace('/^http:\/\//i', 'https://', $redirectUrl);
```
juliushaertl, line 267: This should already be possible to achieve by setting proper reverse proxy headers or or the overwrite protocol config option:

https://docs.nextcloud.com/server/latest/admin_manual/configuration_server/reverse_proxy_configuration.html

I don't think we need an extra setting in user_oidc for that

julien-nc: Very true. I thought this could still be useful in some cases when admins don't have control over the reverse proxy config.

juliushaertl: For that purpose they can set the overwrite protocol ;)

julien-nc: Very true again...sorry for the noise.

juliushaertl: No worries :)

PR #563, opened 2023-01-16
Avoid empty string for scope when inserting/updating

julien-nc:
Because oracle replaces it with null.

We could do it differently by allowing null values for the scope column. Any preferred option?

### DIFF
--- a/lib/Db/Provider.php
+++ b/lib/Db/Provider.php
@@ -63,7 +63,7 @@ class Provider extends Entity implements \JsonSerializable {
 			'identifier' => $this->identifier,
 			'clientId' => $this->clientId,
 			'discoveryEndpoint' => $this->discoveryEndpoint,
-			'scope' => $this->scope,
+			'scope' => trim($this->scope),
 		];
 	}
 }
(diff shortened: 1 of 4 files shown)


lib/Db/Provider.php
```diff
@@ -56,14 +55,21 @@ class Provider extends Entity implements \JsonSerializable {
...
+
 	#[\ReturnTypeWillChange]
 	public function jsonSerialize() {
 		return [
 			'id' => $this->id,
 			'identifier' => $this->identifier,
 			'clientId' => $this->clientId,
 			'discoveryEndpoint' => $this->discoveryEndpoint,
-			'scope' => $this->scope,
+			'scope' => trim($this->scope),
```
juliushaertl, line 72:
```suggestion
			'scope' => trim($this->getScope()),
```

juliushaertl: Could also move the trim inside the getter then ;)

julien-nc: Hmm I think we can't. It seems when updating/inserting a row in the DB, the value that is put there is accessed via the getter. So if we trim there, we will insert/update with an empty string which is what we want to avoid.

juliushaertl: Right, fair enough then

julien-nc: Also, why do you think we should use the getter in jsonSerialize? For empty scopes, the getter will replace it by a space and then we trim it anyway.

PR #580, opened 2023-02-21
Do not expose the expected code when it does not match

julien-nc:
Only expose the expected code when the instance is in debug mode.

### DIFF
--- a/lib/Controller/Id4meController.php
+++ b/lib/Controller/Id4meController.php
@@ -35,7 +35,10 @@ use OCP\AppFramework\Db\DoesNotExistException;
 use OCP\AppFramework\Http;
 use OCP\AppFramework\Http\JSONResponse;
 use OCP\AppFramework\Http\RedirectResponse;
+use OCP\AppFramework\Http\TemplateResponse;
 use OCP\Http\Client\IClientService;
+use OCP\IConfig;
+use OCP\IL10N;
 use OCP\IRequest;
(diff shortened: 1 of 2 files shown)


juliushaertl (approved):

lib/Controller/Id4meController.php
```diff
@@ -173,11 +179,16 @@ public function code($state = '', $code = '', $scope = '') {
...
 		if ($this->session->get(self::STATE) !== $state) {
+			$responseData = [
+				'error' => 'invalid_state',
+				'error_description' => 'received state does not match the expected value',
+			];
+			if ($this->config->getSystemValueBool('debug', false)) {
+				$responseData['got'] = $state;
+				$responseData['expected'] = $this->session->get(self::STATE);
+			}
 			// TODO show page with forbidden
```
nickvergessen, line 190: it's returning a forbidden, so we can remove this? It shows an error in the frontend?

julien-nc: I guess the comment means "returning a proper forbidden page with a nice template".
Now it only shows the raw returned json.

juliushaertl: Maybe we can move to an error template page directly as seeing json in the browser is far from nice ;)

julien-nc: :+1: Done

PR #585, opened 2023-02-27
Fix id4me

julien-nc:
* Fix id4me-related imports
* Catch InvalidOpenIdDomainException exception to show nice error

I tried to import it via mozart but the `src` dir was not added in `lib/Vendor/Id4me`. I didn't wanna fight mozart so missing import are done via requiring composer's autoload file and then we directly import from `vendor`.

### DIFF
--- a/lib/Controller/Id4meController.php
+++ b/lib/Controller/Id4meController.php
@@ -46,6 +46,11 @@ use OCP\IUserManager;
 use OCP\IUserSession;
 use OCP\Security\ISecureRandom;
 
+require_once __DIR__ . '/../../vendor/autoload.php';
+use Id4me\RP\Service;
+use Id4me\RP\Exception\InvalidOpenIdDomainException;
+use Id4me\RP\Model\OpenIdConfig;
+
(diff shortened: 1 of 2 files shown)


lib/Controller/Id4meController.php
```diff
@@ -46,6 +46,11 @@
 use OCP\IUserSession;
 use OCP\Security\ISecureRandom;
```
juliushaertl, line 48: Not a fan of this.

```
diff --git a/composer.json b/composer.json
index 64f65dc..017a998 100644
--- a/composer.json
+++ b/composer.json
@@ -45,7 +45,8 @@
         "classmap_directory": "/lib/autoload/",
         "classmap_prefix": "NEXTCLOUD_USER_OIDC_",
         "packages": [
-            "firebase/php-jwt"
+            "firebase/php-jwt",
+            "id4me/id4me-rp"
         ],
         "delete_vendor_directories": true
     }
```

should make them available through Mozart and then you should be able to use them with `OCA\\UserOIDC\\Vendor\\Id4me...`

Issue #591, opened 2023-03-04
`IUserManager::getDisplayName()` only exists since Nextcloud 25


julien-nc:
https://github.com/nextcloud/user_oidc/blob/main/lib/Service/ProvisioningService.php#L105

TODO:

* [x] fix (#594)
* [x] new release (#596 )

cc @juliushaertl

PR #592, opened 2023-03-04
Do not use `IUserManager::getDisplayName()`

julien-nc:
refs #591

Set max NC version to 24.
Remove `IUserManager::getDisplayName()` usage.


julien-nc:
Replaced by #594

PR #596, opened 2023-03-06
Prepare v1.3.1 release

julien-nc:
Kind of urgent since it fixes some potential crashes with NC <= 24.


juliushaertl:
Thanks for taking care of that

PR #598, opened 2023-03-10
Show redirect URI

### LINKED ISSUE #488: Display Backchannel Logout URL + Redirect URL in the Admin settings
When setting up a new Provider it would be very helpful to have the

- Backchannel Logout URL and
- Redirect URL

displayed in the Admin Settings for the specific provider, to be able to Copy+Paste them directly into the relevant client configurations.

While the Backchannel Logout URL is mentioned in the Documentation, the Redirect URL isn't and it would improve the user experience big time to have them directly available, where the provider is configured.

julien-nc:
closes #488 

image

As mentioned in the issue, knowing the redirect URI helps to configure the client on the provider side.
Authorizing the Nextcloud base URL ending with a wildcard is enough but some admins might want to be stricter.

### DIFF
--- a/src/components/AdminSettings.vue
+++ b/src/components/AdminSettings.vue
@@ -68,10 +68,15 @@
 					:key="provider.id"
 					class="oidcproviders__provider">
 					<div class="oidcproviders__details">
-						<strong>{{ provider.identifier }}</strong><br>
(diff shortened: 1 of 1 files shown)


juliushaertl (approved):
Nice 👍

PR #615, opened 2023-04-21
Improve failure handling in controllers

julien-nc:
Better handling of failures in:

* Id4meController::login
* Id4meController::code
* LoginController::login
* LoginController::code
* LoginController::singleLogoutService
* LoginController::backChannelLogout


lib/Controller/Id4meController.php
```diff
@@ -117,6 +131,8 @@ public function __construct(
 		$this->id4MeMapper = $id4MeMapper;
 		$this->config = $config;
 		$this->l10n = $l10n;
+		$this->loginController = $loginController;
```
nickvergessen, line 134: Injecting a controller into another cries for just creating a shared base class instead 🙈

nickvergessen (approved):

juliushaertl (approved):

PR #616, opened 2023-04-21
Bump max NC version to 27

julien-nc:
And adjust test matrices.


.github/workflows/phpunit.yml
```diff
@@ -37,6 +41,9 @@ jobs:
             server-versions: stable22
           - php-versions: 8.1
             server-versions: stable23
+        include:
```
nickvergessen, line 44: also add 26 + 8.2?

nickvergessen (approved):

juliushaertl (approved):

PR #623, opened 2023-04-26
Make sure the user timezone is set

julien-nc:
This is inspired from what's done in user_saml.
If the timezone is not set for the user in `oc_preferences` (core -> timezone) or if the timezone has never been set by user_oidc (we use the Php session to store that information), we load a script which gets the browser timezone and sends a request to store it.

### DIFF
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -40,5 +40,7 @@ return [
 		['name' => 'Settings#updateProvider', 'url' => '/provider/{providerId}', 'verb' => 'PUT'],
 		['name' => 'Settings#deleteProvider', 'url' => '/provider/{providerId}', 'verb' => 'DELETE'],
 		['name' => 'Settings#setID4ME', 'url' => '/provider/id4me', 'verb' => 'POST'],
+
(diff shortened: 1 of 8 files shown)


juliushaertl:
Makes sense. Just curious did you know how this is done with local user backend and why saml and oidc would need special handling?

julien-nc:
@juliushaertl No I didn't check how the timezone is set/updated on classic login.
I guess you imply that we could check if this can be fixed in the core rather than patching every user backends :grin: Right?

juliushaertl:
Yes, would be nice to just have one generic mechanism. Might not be possible even, but worth a quick look still I'd say.

julien-nc:
So it is done in `OC\Authentication\Login\SetUserTimezoneCommand` https://github.com/nextcloud/server/blob/master/lib/private/Authentication/Login/SetUserTimezoneCommand.php#L45-L56
 which is part of the login chain defined in `OC\Authentication\Login\Chain` https://github.com/nextcloud/server/blob/master/lib/private/Authentication/Login/Chain.php#L103

I don't know where this chain->process is triggered.

juliushaertl:
Only during the regular login post request I fear, doesn't seem that easy to have a shared logic then

juliushaertl (approved):

PR #626, opened 2023-05-05
Set displayName for simple group lists

towo:
Thus far, all groups created from a simple "groups" list that only contains the names of the groups get converted into their IDs, making for a very unintuitive UX.

This tiny modification sets the `displayName` value for the groups to the same name as the source groups, thus preserving the group names while not touching the (generated) group IDs.

### DIFF
--- a/lib/Service/ProvisioningService.php
+++ b/lib/Service/ProvisioningService.php
@@ -333,7 +333,7 @@ class ProvisioningService {
 					$group = $v;
 				} elseif (is_string($v)) {
 					// Handle array of strings, e.g. ["group1", "group2", ...]
-					$group = (object)['gid' => $v];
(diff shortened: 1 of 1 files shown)


towo:
Let's see if the tests fail on this or need to be improved…

julien-nc (approved):
Fine by me.
@MarvinOehlerkingCap What do you think? Any case where this wouldn't be expected?

yavorivanov-cw:
any updates on this one?

jofmi:
hi, is there any update on merging this? :)

julien-nc:
@towo To fix the tests, you need to change line 157 of `tests/unit/Service/ProvisioningServiceTest.php` to:
```
'group2',
```
because this is now the expected display name with this group provisioning data.

nanu-c:
Would be nice to have it working :)

PR #668, opened 2023-08-17
fix: Disable password confirmation for SSO

### LINKED ISSUE #468: Cannot create app passwords due to password prompt
When logged in via a cusom oidc provider (keycloak in my case), I'm prompted for my account password if I try to create an app password. Since I'm not logging in via nextcloud, it's not possible to satisfy this password check.

This seems to be the same issue experienced by the dev of the nextcloud social login plugin: https://github.com/nextcloud/server/issues/9474

Setup:
Fresh install of nextcloud (version 24.0.0),
LDAP user backend plugin enabled and handling user provisioning
OIDC user backend plugin enabled and connected to keycloak. OIDC plugin user id mapped to the ID in ldap, and configured with:
```
  'user_oidc' => [
    'single_logout' => true,
    'auto_provision' => false,
  ],
```

juliushaertl:
Similar behaviour as the user_saml app to avoid asking for password confirmation for SSO where we don't have a password to validate against.

Fixes #468

### DIFF
--- a/lib/Controller/Id4meController.php
+++ b/lib/Controller/Id4meController.php
@@ -295,6 +295,9 @@ class Id4meController extends BaseOidcController {
 		$this->userSession->completeLogin($user, ['loginName' => $user->getUID(), 'password' => '']);
 		$this->userSession->createSessionToken($this->request, $user->getUID(), $user->getUID());
 
+		// Set last password confirm to the future as we don't have passwords to confirm against with SSO
+		$this->session->set('last-password-confirm', strtotime('+4 year', time()));
+
 		return new RedirectResponse(\OC_Util::getDefaultPageUrl());
 	}
 }
--- a/lib/Controller/LoginController.php
+++ b/lib/Controller/LoginController.php
@@ -502,6 +502,9 @@ class LoginController extends BaseOidcController {
 		$this->userSession->createSessionToken($this->request, $user->getUID(), $user->getUID());
 		$this->userSession->createRememberMeToken($user);
 
+		// Set last password confirm to the future as we don't have passwords to confirm against with SSO
+		$this->session->set('last-password-confirm', strtotime('+4 year', time()));
+
 		// for backchannel logout
 		try {
 			$authToken = $this->authTokenProvider->getToken($this->session->getId());


thlehmann-ionos:
I don't think this is sufficient and completely thought through.

**TL;DR:** The password confirmation library also checks for [`backendAllowsPasswordConfirmation`][git-nclib-ncpassconf-main], which should be `false`, since a user authenticated via SSO can not confirm their password via Nextcloud's user database. The variable however [will only be `false` if the user backend implements `IPasswordConfirmationBackend`][git-nc-jscfghelper].

## More details

This can only be the case if the user could be looked up in the user_oidc backend by their ID. See user [`Manager->get()`][git-nc-usermgr-get].

This, in turn, means the user has to be provisioned in the user_oidc backend by the same ID as in the Nextcloud backend, since otherwise, the backend lookup would yield the Nextcloud user backend, which does not implement the above mentioned interface. See how the user is looked up in the [code auth phase][git-uo-login-code-userlookup] and later [associated with the user's session][git-uo-login-code-sesssetuser].

The error would be reproducible if the user was auto-provisioned and the (`--unique-id`) was _not_ `false` (`0`) [(`true` is the default)][git-uo-readme-uniqueid] since user_oidc would [generate a new ID][gut-uo-usermapper] in the process. [See the code][git-uo-login-code-autoprov].

Note that a auto-provisioned user _with unique user ID_ could not be looked up as the backend implementing `IPasswordConfirmationBackend` as it would by looked up by the ID in the sub claim. This is a separate issue #898.


## Proposed solutions

1. Hard (=non-auto) de-provisioning should be implemented to allow an explicitly created user_oidc backed user entity to be removed, since it otherwise breaks de-provisioning via Nextcloud too or
2. The hard provisioning should provision the user in the Nextcloud backend
3. The ID-uniqueness feature should be dropped or changed in a way that the user_oidc backend allows lookup by the hashed user ID too



[git-nclib-ncpassconf-main]: https://github.com/nextcloud-libraries/nextcloud-password-confirmation/blob/v5.1.1/src/main.ts#L21
[git-nc-jscfghelper]: https://github.com/nextcloud/server/blob/v29.0.3/lib/private/Template/JSConfigHelper.php#L89
[git-nc-usermgr-get]: https://github.com/nextcloud/server/blob/v29.0.3/lib/private/User/Manager.php#L157
[git-uo-login-code-userlookup]: https://github.com/nextcloud/user_oidc/blob/v5.0.3/lib/Controller/LoginController.php#L510
[git-uo-login-code-sesssetuser]: https://github.com/nextcloud/user_oidc/blob/v5.0.3/lib/Controller/LoginController.php#L539
[git-uo-readme-uniqueid]: https://github.com/nextcloud/user_oidc/blob/v5.0.3/README.md#user-ids
[gut-uo-usermapper]: https://github.com/nextcloud/user_oidc/blob/v5.0.3/lib/Db/UserMapper.php#L133
[git-uo-login-code-autoprov]: https://github.com/nextcloud/user_oidc/blob/v5.0.3/lib/Controller/LoginController.php#L524

PR #677, opened 2023-08-31
Telekom bearer token: handling

tsdicloud:
The branch adds the handling of Telekom (encrypted, signed) JWT bearer token to user_oidc 
replacing the standard Openid bearer token handling with own `MBackend`.


tsdicloud:
Sorry, wrong project

PR #688, opened 2023-09-27
Fix for Telekom backchannel logout

tsdicloud:
We needed to switch back to Nextcloud delivered version for getting the secured version and profit from future updates.

Making it work for Telekom needs some fixes (some of that need to be contributed upstream).


tsdicloud:
Alltime classics: sorry, wrong project.

PR #697, opened 2023-10-27
Add PKCE support

### LINKED ISSUE #344: Support for PKCE
https://datatracker.ietf.org/doc/html/rfc7636

There are some openid connect providers that required PKCE to function correctly, which it appears that this plugin doesn't support in it's requests. It would be great if this could be updated to have PKCE S256 functionality. 

For more: https://medium.com/swlh/pkce-flow-of-openid-connect-9b10ddbabd66 

Thanks!

rullzer:
For the providers that support it. Add extra security. See also https://datatracker.ietf.org/doc/html/rfc7636

closes #344 #710


nc-fkl (approved):
thank you @rullzer for your commit, i fixed a typo and rebased on main

PR #711, opened 2023-11-21
increased column length for id and secret

### LINKED ISSUE #405: Client ID and Client Secret string limits are too short.
My Nextcloud implementation works with OpenID Connect (OIDC) via Keycloak but not with OIDC via Google or Paypal. The two platforms create the same issue when I register the provider:

```
sudo -u www-data php /var/www/nextcloud/occ user_oidc:provider google --clientid="8********************************************.apps.googleusercontent.com" --clientsecret="G**********************************" --discoveryuri="https://accounts.google.com/.well-known/openid-configuration"
```

The command returns this error on my server:

```
In DbalException.php line 71:An exception occurred while executing a query: SQLSTATE[22001]: String data, right truncated: 1406 Data too long for column 'client_id' at row 1

In ExceptionConverter.php line 114: An exception occurred while executing a query: SQLSTATE[22001]: String data, right truncated: 1406 Data too long for column 'client_id' at row 1

In Exception.php line 26: SQLSTATE[22001]: String data, right truncated: 1406 Data too long for column 'client_id' at row 1

In Statement.php line 92: SQLSTATE[22001]: String data, right truncated: 1406 Data too long for column 'client_id' at row 1
```

The Google client secret has this format: `8********************************************.apps.googleusercontent.com`. Each star represents a unique character. I considered dropping the `.apps.googleusercontent.com` extension, but Google includes the suffix [in their own examples](https://developers.google.com/identity/protocols/oauth2/openid-connect#sendauthrequest): `client_id=424911365001.apps.googleusercontent.com`.

Paypal has the same problem with Client ID. The service also uses a longer Client Secret. So I get more errors with that service.

nc-fkl:
closes #405
Increased column limit to 2048 for client_secret and client_id as there is no specified max length in https://www.rfc-editor.org/rfc/rfc6749

### DIFF
--- a/lib/Migration/Version010304Date20231121102449.php
+++ b/lib/Migration/Version010304Date20231121102449.php
@@ -1,0 +1,54 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OCA\UserOIDC\Migration;
+
+use Closure;
+use OCP\DB\ISchemaWrapper;
+use OCP\IDBConnection;
+use OCP\Migration\IOutput;
+use OCP\Migration\SimpleMigrationStep;
+use OCP\Security\ICrypto;
+
+class Version010304Date20231121102449 extends SimpleMigrationStep {
+
+	/**
+	 * @var IDBConnection
+	 */
+	private $connection;
+	/**
+	 * @var ICrypto
+	 */
+	private $crypto;
+
+	public function __construct(
+		IDBConnection $connection,
+		ICrypto $crypto
+	) {
+		$this->connection = $connection;
+		$this->crypto = $crypto;
+	}
+
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options) {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		foreach (['user_oidc_providers', 'user_oidc_id4me'] as $tableName) {
+			if ($schema->hasTable($tableName)) {
+				$table = $schema->getTable($tableName);
+				if ($table->hasColumn('client_secret')) {
+					$column = $table->getColumn('client_secret');
+					$column->setLength(2048);
+					$column = $table->getColumn('client_id');
+					$column->setLength(2048);
+					$schemaChanged = true;
+				}
+			}
+		}
+		if ($schemaChanged) {
+			return $schema;
+		}
+		return null;
+	}
+}
--- a/lib/Migration/Version010304Date20231121102449.php
+++ b/lib/Migration/Version010304Date20231121102449.php
@@ -34,6 +34,7 @@ class Version010304Date20231121102449 extends SimpleMigrationStep {
 		/** @var ISchemaWrapper $schema */
 		$schema = $schemaClosure();
 
+		$schemaChanged = false;
 		foreach (['user_oidc_providers', 'user_oidc_id4me'] as $tableName) {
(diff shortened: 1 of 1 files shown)


lib/Migration/Version010304Date20231121102449.php
```diff
@@ -0,0 +1,53 @@
...
+
+		foreach (['user_oidc_providers', 'user_oidc_id4me'] as $tableName) {
+			if ($schema->hasTable($tableName)) {
+				$table = $schema->getTable($tableName);
+				if ($table->hasColumn('client_secret')) {
+					$column = $table->getColumn('client_secret');
+					$column->setLength(2048);
+					$column = $table->getColumn('client_id');
+					$column->setLength(2048);
+					return $schema;
```
julien-nc, line 46: This will return the schema on the first loop iteration so the change won't be made in `user_oidc_id4me`.
You can set a flag here instead (`$schemaChanged = true;`) and then after the loop:
``` php
if ($schemaChanged) {
    return $schema;
}
```

lib/Migration/Version00004Date20200428102743.php
```diff
@@ -37,11 +37,11 @@ public function changeSchema(IOutput $output, Closure $schemaClosure, array $opt
 		]);
 		$table->addColumn('client_id', 'string', [
 			'notnull' => true,
-			'length' => 64,
+			'length' => 2048,
```
julien-nc, line 40: For fresh installations of the app, all the migration steps will run so it's not necessary to change the old one.

julien-nc (changes requested):
Nice!
It would be nice to show the max length in the settings labels and to limit the input fields.

julien-nc:
:ping_pong: :grin:

lib/Migration/Version010304Date20231121102449.php
nc-fkl: done :)

nc-fkl:
> 🏓 😁

🏓pong

lib/Migration/Version010304Date20231121102449.php
```diff
@@ -0,0 +1,54 @@
...
+		foreach (['user_oidc_providers', 'user_oidc_id4me'] as $tableName) {
+			if ($schema->hasTable($tableName)) {
+				$table = $schema->getTable($tableName);
+				if ($table->hasColumn('client_secret')) {
+					$column = $table->getColumn('client_secret');
+					$column->setLength(2048);
+					$column = $table->getColumn('client_id');
+					$column->setLength(2048);
+					$schemaChanged = true;
+				}
```
julien-nc, line 46:
```suggestion
				if ($table->hasColumn('client_secret')) {
					$column = $table->getColumn('client_secret');
					$column->setLength(2048);
					$schemaChanged = true;
				}
				if ($table->hasColumn('client_id')) {
					$column = $table->getColumn('client_id');
					$column->setLength(2048);
					$schemaChanged = true;
				}
```

Also, schemaChanged is not initialized (to false).

julien-nc (changes requested):

lib/Migration/Version010304Date20231121102449.php
julien-nc: schemaChanged is still not initialized :grin:

julien-nc (approved):

PR #713, opened 2023-11-21
Fix potentially missing "alg" in jwks

### LINKED ISSUE #709: regression: JWK must contain an alg parameter
**Steps to reproduce**

    * update user_oidc from 1.3.3 to 1.3.4

**Expected behavior**

  * login still works as usual

**Actual behaviour**

After updating login via oidc is not possible any more. Error:  
```
  "Exception": "UnexpectedValueException",
   "Message": "JWK must contain an "alg" parameter",
```

Downgrading user_oidc to 1.3.3. fixes the issue. 

It seems my oidc idp (shibboleth with oidc plugin) does not supply this param and according to [rfc-7517](https://datatracker.ietf.org/doc/html/rfc7517#section-4.4) alg is optional

> The "alg" value is a case-sensitive ASCII string.  Use of this member is OPTIONAL.

Help please

julien-nc:
closes #709 

Some OpenID providers do not include an "alg" key in their list of JWKs. This was not an issue but with php-jwt now requires an "alg" in a JWK to decode a token.

This fix gets the alg from the token we are trying to decode and puts it in the JWK array before php-jwt parses it.

### DIFF
--- a/lib/Controller/LoginController.php
+++ b/lib/Controller/LoginController.php
@@ -414,7 +414,7 @@ class LoginController extends BaseOidcController {
 
 		// TODO: proper error handling
 		$idTokenRaw = $data['id_token'];
-		$jwks = $this->discoveryService->obtainJWK($provider);
+		$jwks = $this->discoveryService->obtainJWK($provider, $idTokenRaw);
 		JWT::$leeway = 60;
 		$idTokenPayload = JWT::decode($idTokenRaw, $jwks);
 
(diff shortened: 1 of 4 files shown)


juliushaertl (approved):
Not sure how to test but the code looks sane 👍

julien-nc:
Thanks for the review. I tested it by manually removing the alg attr of the JWK array received from the discovery endpoint.

Fabien4941:
Thanks, I had the same error. I tried your modifications and It works with Azure.

Thansk again for your work !

PR #714, opened 2023-11-21
Disabled the possibility to edit the identifier

nc-fkl:
Disabled the possibility to edit the identifier when its already set and saved

### DIFF
--- a/src/components/SettingsForm.vue
+++ b/src/components/SettingsForm.vue
@@ -20,10 +20,13 @@
   -
   -->
 
-  <template>
+<template>
 	<form class="provider-edit">
 		<p>
+			<!--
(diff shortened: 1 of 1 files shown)


julien-nc:
Nice!

It would be nice to also display the max identifier size when it can be edited (on provider creation). It is 128 characters. The input field should be limited to 128 chars too.

In a future PR we could add a display name field that would always be editable

julien-nc:
Oh and apparently your commit signatures are incorrect somehow.

julien-nc (changes requested):
:+1: A few remarks.
I think it's nice to display the identifier max size in the related label string.
Something like `t('user_oidc', 'Identifier (max 128 chars')`.

julien-nc (approved):
:+1:

PR #716, opened 2023-11-23
chore(CI): Adjust testing matrix for Nextcloud 28 on main

nickvergessen:
Signed-off-by: Joas Schilling <coding@schilljs.com>


julien-nc (approved):

julien-nc:
We also need to bump the max NC version.

nickvergessen:
> We also need to bump the max NC version.

That comes in a second PR usually:
https://github.com/nextcloud/user_oidc/pull/719

Issue #633, opened 2023-05-21
Auhentik and user_oidc:  "Wrond number of segments" errors in the log


Ra72xx:
I switched Nextcloud from LDAP auth to OIDC according to this page: [https://blog.cubieserver.de/2022/complete-guide-to-nextcloud-oidc-authentication-with-authentik/](url). I also shortened my token because this is mentioned in this guide. Though everything seems to work as expected, I get tons of messages like this in my Nextcloud log:
`
{"reqId":"xxxxx","level":3,"time":"2023-05-21T04:31:00+00:00","remoteAddr":"46.142.186.191","user":"--","app":"user_oidc","method":"GET","url":"/nextcloud/index.php/apps/passwords/api/1.0/session/keepalive","message":"Impossible to decode OIDC token:Wrong number of segments","userAgent":"Offizielle Passwords Firefox Extension auf Linux","version":"26.0.1.1","data":{"app":"user_oidc"},"id":"xxxxxx"}`

This also happens for other apps, not only the passwords extension. However, as I said, everything seems to work, but I dislike logfile entries like this ;-) ...

quenenni:
Hello,

I have the same situation.
From what I gathered, the log message comes when you're using a app password token instead of a SSO token to access the cloud from an external app.

That msg comes from this file:
apps/user_oidc/lib/Vendor/Firebase/JWT/JWT.php

So, indeed, that shouldn't be a problem, but maybe the dev can check if it's an app passwrod token used and in that case, not show the msg.

joergmschulz:
this is not restricted to Authentik but happens in keycloak as well.

joshtrichards:
Fixed in #721

PR #721, opened 2023-11-24
fix error: if Bearer token is no JWT

aro-lew:
possible fix for #633, since only JWTs need further analyze.


aro-lew:
Yeah, that would be good for me.
I can't imagine a scenario where `JWT::decode` in `lib/User/Validator/SelfEncodedValidator.php` throws an error that should be in the error log.
I will update my PR and also set the token expired message to debug.

to-kn:
@julien-nc any eta when there will be a new release?

PR #724, opened 2023-11-30
Added customizeable end session endpoint to frontend and backend

### LINKED ISSUE #599: Feature: Add option to set logout URL manually
Would it be possible to add a configuration option, either in the UI or in the config file, to manually set the logout URL? The OpenID provider I am using, Authelia, doesn't seem to provide a `end_session_endpoint` value in its `/.well-known/openid-configuration` response.

Later edit: But they do have a logout page, which if redirected to, serves the purpose. If we could redirect to that page, it would allow us to log out of SSO from within Nextcloud.

nc-fkl:
closes #599 
- New optional End session endpoint field to the settings frontend
- Added new end session endpoint column via migration to the database table
- Check in the backend if custom end session endpoint is stored and manipulate the discovery endpoint response if its true


lib/Service/DiscoveryService.php
```diff
@@ -80,6 +80,16 @@ public function obtainDiscovery(Provider $provider): array {
...
+
+			// Manipulate the response with the custom endpoint url
+			$endpointData = json_encode($this->providerService->getProviderWithSettings($provider->getId()));
+			$endpointData = json_decode($endpointData)->endSessionEndpoint;
+			if ($endpointData) {
+				$discoveryData = json_decode($cachedDiscovery);
+				$discoveryData->end_session_endpoint = $endpointData;
+				$cachedDiscovery = json_encode($discoveryData);
+			}
+
```
julien-nc, line 92: Replacing/setting the custom end session endpoint in the discovery itself means it is being cached so changing it won't be effective immediately.
Since this value is only used in `LoginController::singleLogoutService`, I think it's better to access it there, directly from the provider object.
Same logic as what you did here: Use the custom value if it's set, fallback on the one from the discovery.

Small detail (not useful here anymore): `json_decode` has a second `associative` boolean parameter. Setting it to true makes the function return an associative array instead of an object. We usually do it this way.

lib/Db/Provider.php
```diff
@@ -36,6 +36,7 @@
  * @method void setClientSecret(string $clientSecret)
  * @method string getDiscoveryEndpoint()
  * @method void setDiscoveryEndpoint(string $discoveryEndpoint)
+ * @method void setEndSessionEndpoint(string $endSessionEndpoint)
```
julien-nc, line 39: We need `getEndSessionEndpoint` as well.

src/components/SettingsForm.vue
```diff
@@ -231,6 +239,9 @@ export default {
 			min-width: 200px;
 			flex-grow: 1;
 		}
+		input[type=text].italic-placeholder::placeholder {
```
julien-nc, line 242: Nitpicking :grin: 
```suggestion
		.italic-placeholder::placeholder {
```

lib/Migration/Version010304Date20231130104459.php
```diff
@@ -0,0 +1,56 @@
...
+	 */
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		$table = $schema->getTable('user_oidc_providers');
+		$table->addColumn('end_session_endpoint', 'string', [
+			'notnull' => false,
+			'length' => 255,
+		]);
```
julien-nc, line 52:
```suggestion
		if ($schema->hasTable('user_oidc_providers')) {
			$table = $schema->getTable('user_oidc_providers');
			if (!$table->hasColumn('end_session_endpoint')) {
				$table->addColumn('end_session_endpoint', Types::STRING, [
					'notnull' => false,
					'length' => 255,
				]);
			}
		}
```
* Let's stay defensive in migration steps :wink: 
* The column types can be found as constants in the `OCP\DB\Types` class

julien-nc (changes requested):
Nice!

lib/Service/DiscoveryService.php
nc-fkl: Fixed - moved it into the login controller and adjusted the code

lib/Db/Provider.php
nc-fkl: Added now - thanks for the suggestion! :)

lib/Migration/Version010304Date20231130104459.php
nc-fkl: Good point - adjusted it

lib/Migration/Version010304Date20231130104459.php
```diff
@@ -0,0 +1,61 @@
...
+			$table = $schema->getTable('user_oidc_providers');
+			if (!$table->hasColumn('end_session_endpoint')) {
+				$table->addColumn('end_session_endpoint', Types::STRING, [
+					'notnull' => false,
+					'length' => 255,
+				]);
+			}
+		}
+
+		return $schema;
```
julien-nc, line 59:
```suggestion
		if ($schema->hasTable('user_oidc_providers')) {
			$table = $schema->getTable('user_oidc_providers');
			if (!$table->hasColumn('end_session_endpoint')) {
				$table->addColumn('end_session_endpoint', Types::STRING, [
					'notnull' => false,
					'length' => 255,
				]);
				return $schema;
			}
		}

		return null;
```
Null should be returned if the schema has not been changed.

julien-nc (changes requested):
Sorry for the late review.

julien-nc (changes requested):
One last thing, it would be nice if the custom end_session endpoint could be set in the Upsert command.

nc-fkl:
> One last thing, it would be nice if the custom end_session endpoint could be set in the Upsert command.

Its also in now  :raised_hands:
Can you please re-review? :ping_pong:

julien-nc (approved):
I added a few adjustments.
* set max input field length in the UI
* don't try to use the custom endpoint if it's an empty string
* allow missing endSessionEndpoint param in SettingsController::createProvider
* update a doc block

Issue #656, opened 2023-08-01
Nextcloud + Keycloak + Lowercase


brtptrs:
Hi,

we've decided to switch our NC authentication from username+password to openid-conect using Keycloak and user_oidc

Everything seemed to work fine but:
1) In NC many of our usernames contain uppercase letters
2) Keycloak converts all usernames to lowercase
3) For users with uppercase usernames user_oidc creates new duplicate, lowercase accounts, instead of logging into the existing accounts
4) Renaming users in NC is almost impossible since we have many users with lots of files

Can this behavior be changed?

brtptrs:
We've hacked away at the provisioning code and have come up with a solution that seems to work for us.
Since we usually don't do Nextcloud coding, it would be great if someone could take a look at the patch, so see if it contains any major blunders.

[camelcase-patch.zip](https://github.com/nextcloud/user_oidc/files/12259891/camelcase-patch.zip)

PR #730, opened 2023-12-08
Implement soft auto provisioning

julien-nc:
Problem: When auto provisioning is enabled, if a user with the same ID exists in another backend (Db, LDAP...), we create a second user in the user_oidc backend which leads to issues and confusion.

Here is a partial solution to this problem:
New config `soft_auto_provision` flag (enabled by default).
* When enabled
    * If the user already exists in another backend, we don't create a new one in our backend. We update the information (mapped attributes) of the existing user.
    * If the user does not exist in another backend, we create it in ours
* When disabled
    * We refuse login of users that already exist in other backends

This does not migrate existing users to the user_oidc backend but allows them to log in without creating a duplicate user.

Nothing was changed when auto provisioning is disabled.

Extra change: In the user backend, the ldap sync was not triggered when looking for an existing user.


lib/Service/ProvisioningService.php
```diff
@@ -93,17 +95,21 @@ public function provisionUser(string $tokenUserId, int $providerId, object $idTo
...
+					$this->userMapper->update($backendUser);
+				}
+				// 2 reasons why we should update the display name: It does not match the one
+				// - of our backend
+				// - returned by the user manager (outdated one before the fix in https://github.com/nextcloud/user_oidc/pull/530)
+				if ($newDisplayName !== $oldDisplayName || $newDisplayName !== $user->getDisplayName()) {
+					$this->eventDispatcher->dispatchTyped(new UserChangedEvent($user, 'displayName', $newDisplayName, $oldDisplayName));
+				}
+			} else {
+				$user->setDisplayName($newDisplayName);
```
nc-fkl, line 112: Not sure here if the $newDisplayName in L112 is always set as its only set once when the if condition in L101 was triggered

nc-fkl (changes requested):
Please check my comment regarding $newDisplayName in ProvisioningService.php

nc-fkl (approved):
Nice! :)

julien-nc:
@juliushaertl Could you check if this PR makes sense? :grin:

juliushaertl (approved):
The approach looks good to me, haven't tested though 👍

brtptrs:
Would it be possible to catch this Problem during while doing this provisioning?
https://github.com/nextcloud/user_oidc/issues/656
Maybe with a config flag to enable a lowercase - uppercase verification

PR #733, opened 2023-12-12
Implement ICountUsersBackend to give a user count in 'occ user:report'

julien-nc:
And fix double user backend registration.
@juliushaertl Do you know why the double registration was there in the first place?

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -62,7 +62,9 @@ class Application extends App implements IBootstrap {
 
 		/* Register our own user backend */
 		$this->backend = $this->getContainer()->get(Backend::class);
-		$userManager->registerBackend($this->backend);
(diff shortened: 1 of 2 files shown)


nc-fkl (approved):

Rello:
Bildschirmfoto 2023-12-12 um 16 48 52

category is showing up now

Rello (approved):

juliushaertl:
No idea, in general i think it is odd that there are two different ways for that. OC_User needs to be used for the apache backend handling in https://github.com/nextcloud/server/blob/ea8f9a7e841061ed96f291aa3c9a413fe2402b5c/lib/private/legacy/OC_User.php#L410, should probably be reworked on the server at some point to actually have that in IRegistrationContext, but the PR here seems fine for now 👍

PR #740, opened 2023-12-21
Make pkce optional

### LINKED ISSUE #735: Error after updating to 1.3.5
Steps to reproduce

update from 1.3.3 to 1.3.5

Expected behavior

OPENID Login page shown
Actual behaviour

When visiting the page i get this error:

```
{
    "error": "invalid_request",
    "error_description": "Invalid request"
}
```
Downgrade to 1.3.3 was a workaround but after the update to 28.0.0.0 we cannot downgrade this app anymore.

There is no log in Nextcloud.

julien-nc:
closes #735

Some providers like Entrust refuse to show the login page and return those to the code endpoint when PKCE parameters are passed in the login page URL: `{"error":"invalid_request","error_description":"Invalid request"}`.

This disables PKCE by default. It can be enabled in config.php

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,17 @@ parameter to the login URL.
 sudo -u www-data php var/www/nextcloud/occ config:app:set --value=0 user_oidc allow_multiple_user_backends
 ```
 
+### PKCE
+
+This app supports PKCE (Proof Key for Code Exchange).
+https://datatracker.ietf.org/doc/html/rfc7636
+It is disabled by default and can be enabled in `config.php`:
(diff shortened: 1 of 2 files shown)


nc-fkl (approved):
looks good to me

Kranzes:
PKCE should be enabled by default, PKCE is the right way to do OpenID Connect according to the spec. Most Idp's support PKCE I don't think it makes sense to disable PKCE by default because of one flawed ldp. The recent update broke the login functionality for most people I assume.

PR #750, opened 2024-01-04
Ensure the discovery endpoint is validated and all required fields ar…

nc-fkl:
Ensure the discovery endpoint is validated and all required fields are verified

### DIFF
--- a/lib/Controller/SettingsController.php
+++ b/lib/Controller/SettingsController.php
@@ -63,10 +63,57 @@ class SettingsController extends Controller {
 		$this->crypto = $crypto;
 	}
 
+	public function isDiscoveryEndpointValid($url) {
+		// Initialize cURL session
+		$ch = curl_init($url);
+		curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
+		curl_setopt($ch, CURLOPT_HEADER, false);
+		$response = curl_exec($ch);
+		$httpCode = curl_getinfo($ch, CURLINFO_HTTP_CODE);
+		curl_close($ch);
+	
+		$result = [
+			'isReachable' => false,
+			'missingFields' => [],
+		];
+	
+		// Check if the request was successful
+		if ($httpCode == 200 && !empty($response)) {
+			$result['isReachable'] = true;
+			$data = json_decode($response, true);
+	
+			// Check for necessary fields in the response
+			$requiredFields = [
+				'issuer', 'authorization_endpoint', 'token_endpoint', 'jwks_uri',
+				'response_types_supported', 'subject_types_supported', 'id_token_signing_alg_values_supported',
+			];
+			
+			foreach ($requiredFields as $field) {
+				if (!isset($data[$field])) {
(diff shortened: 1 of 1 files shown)


lib/Controller/SettingsController.php
```diff
@@ -63,12 +63,59 @@ public function __construct(
 		$this->crypto = $crypto;
 	}
 
+	public function isDiscoveryEndpointValid($url) {
+		// Initialize cURL session
```
juliushaertl, line 67: This should use https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/http_client.html, not plain curl

lib/Controller/SettingsController.php
```diff
@@ -47,20 +49,63 @@ class SettingsController extends Controller {
...
+				
+				foreach ($requiredFields as $field) {
+					if (!isset($data[$field])) {
+						$result['missingFields'][] = $field;
+					}
+				}
+			} else {
+				// Set isReachable to false if http code wasn't 200
+				$result['isReachable'] = false;
+			}
```
juliushaertl, line 103:
```suggestion
			}
```

No need for this else branch as isReachable is only set to true if the status code is 200 in line 86

lib/Controller/SettingsController.php
```diff
@@ -47,20 +49,63 @@ class SettingsController extends Controller {
...
+		} catch (Exception $e) {
+			$result['isReachable'] = false;
```
juliushaertl, line 105: Same here, but still worth to catch and best log the exception here

juliushaertl (commented):
Two additional minor comments, otherwise this looks sane :)

lib/Controller/SettingsController.php
```diff
@@ -47,20 +50,64 @@ class SettingsController extends Controller {
...
+		];
+
+		try {
+			$client = $this->clientService->newClient();
+			$response = $client->get($url);
+			$httpCode = $response->getStatusCode();
+			$body = $response->getBody();
+
+			// Check if the request was successful
+			if ($httpCode == 200 && !empty($body)) {
```
julien-nc, line 90:
```suggestion
			if ($httpCode === Http::STATUS_OK && !empty($body)) {
```
Use triple equal and use `OCP\AppFramework\Http` constants for HTTP response codes.

lib/Controller/SettingsController.php
```diff
@@ -47,20 +50,64 @@ class SettingsController extends Controller {
...
+				];
+				
+				foreach ($requiredFields as $field) {
+					if (!isset($data[$field])) {
+						$result['missingFields'][] = $field;
+					}
+				}
+			}
+		} catch (Exception $e) {
+			$this->logger->error('Error - discovery endpoint validation: ' . $e);
```
julien-nc, line 107:
```suggestion
			$this->logger->error('Discovery endpoint validation error', ['exception' => $e]);
```
So we get more info on the exception.

Issue #743, opened 2023-12-28
"csrf check failed" after Nextcloud session ends


isdnfan:
# Problem
I successfully integrated Nextcloud with Zitadel IdP using user_oidc but I hit an issue with **allow_multiple_user_backends=0** config.

# Setup
The idea was to reduce Nextcloud session lifetime so NC session ends quickly and the user must re-login using IdP to ensure user session is still valid in IdP. To ensure I configure following settings in NC:

| setting | value |
|---|---|
|auto_logout|false|
|session_keepalive|true|
|session_lifetime|120|
|session_relaxed_expiry|false|
|remember_login_cookie_lifetime|0|

- session lifetime could be longer I started with 15min, such extremely short value is used to hit the issue fast.

with  **allow_multiple_user_backends=1**  the settings work fine and the user returns to login screen where hitting the button "login with IdP" allows to start another session. 

![image](https://github.com/nextcloud/user_oidc/assets/18125597/2c1acfb0-cc7f-47da-b189-e220999c0b98)

the problem starts when I forced IdP login **allow_multiple_user_backends=0** using `occ config:app:set --value=0 user_oidc allow_multiple_user_backends` which worked as expected immediately redirecting unauthorized user to IdP and allowing access upon successful authorization. But after Nextcloud session ends the user is unable to return to Nextcloud. The browser keeps bouncing between Nextcloud and IdP with requests 
- Nextcloud/logout?requesttoken=123
- IdP/authorize
- Nextcloud/login?redirect_url=/logout?requesttoken=123.. 

keeping requesttoken constant and at some point hitting 412 "CSRF check failed"

![image](https://github.com/nextcloud/user_oidc/assets/18125597/29ddc374-3954-4b66-89a6-04faabd853cc)

# How to reproduce
- setup Nextcloud user_oidc and some IdP
- reduce very short NC session timeout
- login using a browser (in my case Firefox on Windows)
- default view in my instance is files app
- do nothing and let the session open
- using F12 tools permanent exchange of push/sync messages is visible
- after Nextcloud session ends the browser starts looping between NC and IdP

# Logs
I'm adding anonymized HAR file from browser dev tools showing the issue. In this log *https://dev-nc.mydomain.tld* is my Nextcloud and *https://sso.mydomain.tld* is the IdP. In my case I'm using Zitadel but the same issue happens with authentik and Keycloak as well.

[...]

PR #761, opened 2024-01-16
Fix potential login -> logout loop when allow_multiple_user_backends is 0

julien-nc:
This makes sure the user_oidc login page does not allow the `redirect_url` GET param to be the logout page or `/apps/user_oidc/sls` (single logout) when allow_multiple_user_backends == 0.

The issue is well described in #743. I could reproduce it with those system setting values:
```
'auto_logout' => true,
'session_lifetime' => 30,
'session_keepalive' => false,
'remember_login_cookie_lifetime' => 35,
```
Then authenticate via user_oidc, then wait for the auto logout.

@isdnfan Could you check if it makes things better on your side?

### DIFF
--- a/lib/Controller/LoginController.php
+++ b/lib/Controller/LoginController.php
@@ -209,6 +209,16 @@ class LoginController extends BaseOidcController {
 	 * @return DataDisplayResponse|RedirectResponse|TemplateResponse
 	 */
 	public function login(int $providerId, string $redirectUrl = null) {
+		// to be safe, avoid redirecting to logout or single-logout
+		$logoutUrl = $this->urlGenerator->linkToRoute('core.login.logout');
+		$userOidcLogoutUrl = $this->urlGenerator->linkToRoute(Application::APP_ID . '.login.singleLogoutService');
+		if (
+			$redirectUrl
+			&& (strpos($redirectUrl, $logoutUrl) !== false || strpos($redirectUrl, $userOidcLogoutUrl) !== false)
+		) {
+			$redirectUrl = $this->urlGenerator->getBaseUrl();
+		}
+
 		if ($this->userSession->isLoggedIn()) {
 			return new RedirectResponse($redirectUrl);
 		}


julien-nc:
@isdnfan Answering to https://github.com/nextcloud/user_oidc/issues/743#issuecomment-1894537506 here if you don't mind :grin: .

Could you try again? There was a bug in the fix :see_no_evil: . It's now fixed.

The auto redirection didn't work when the `redirect_url` GET param of the login page was not set. Big side effect: the crash prevented the "alternative login" button to be registered so even when reaching the login page (after the failed redirect), there was no button to manually go to the IdP. All should be fine now.

isdnfan:
unfortunately the loop is back with he new version:

image

julien-nc:
The loop that is addressed here is the one started when auto_logout is true and user_oidc's `allow_multiple_user_backends` is 0.
With those config values:
```
'auto_logout' => true,
'session_lifetime' => 30,
'session_keepalive' => false,
'remember_login_cookie_lifetime' => 35,
```
image
When the user is auto-logged out, there's a redirection to `login?redirect_url=/dev/server/index.php/logout?requesttoken%3DCY8Cx14ys0exU...` which was leading to a loop but now it's redirected to `/apps/user_oidc/login/3?redirectUrl=https://nextcloud.local/dev/server`. The `redirectUrl` GET param of the user_oidc login cannot be the logout page anymore.

I tried setting the config values you mentioned in the issue:
```
'auto_logout' => false,
'session_keepalive' => true,
'session_lifetime' => 30,
'session_relaxed_expiry' => false,
'remember_login_cookie_lifetime' => 0,
```
but could not enter any loop (with or without the changes). Could you give a precise description of what you do to enter the loop? I need all the steps to be able to reproduce. Also, might not be relevant but: Are you using a chromium based browser?
According to your screenshot, I don't understand how the loop is started because the logout redirect seems to be correctly replaced by your instance base URL
image

isdnfan:
hi Julien I'm using Firefox 121.0.1 on Windows 10 (but loop happens on Edge as well). I'm doing "nothing" to enter the loop. In my installation deck app is disabled so main app is files 
- I just visit my main page 'https://dev-nc.mydomain.tld' 
- redirect to IdP happens 
- after login I see "Files" app
- in dev tools every 5 sec there is push/sync message send back and forth
- after session timeout at 30-35 sec I see 
- `GET /logout?requesttoken=...`

I was under impression the issue is related to 'auto_logout' after your last update it worked when the setting was true and looped at false. But switching it back and was not really stable. Additionally I observer an interesting behavior: with the last fix the loop happens but after looping for 10 or 20 times there successful login happens and the browser hits the Files app page.. 

For me 'GET /login?redirect_url=/logout?requesttoken' sounds strange - if I get it right this makes the client logs just to proceed with logout again? this request is the first request when the client comes back from IdP. I added the whole browser HAR file - dev-nc. is Nextcloud 28.0.1 and sso. is IdP (Zitadel) - just replaced my domain with <mydomain.tld>. 

[dev-nc.mydomain.tld_Archive [24-01-17 13-22-52].har.zip](https://github.com/nextcloud/user_oidc/files/13963253/dev-nc.mydomain.tld_Archive.24-01-17.13-22-52.har.zip)

julien-nc:
@isdnfan Thanks. I need more details to make sure i'm in the same context:
* config.php values
* what's the target URL of the push/sync requests?
* what exactly do you see when the loop starts? Is the logout request done by a script or is the entire page redirected to /logout?

isdnfan:
hi @julien-nc still no success :(

a don't spot any difference to the previous state.. I add the new HAR file as well 

[dev-nc.mydomain.tld_Archive [24-01-18 21-55-39].har.zip](https://github.com/nextcloud/user_oidc/files/13981639/dev-nc.mydomain.tld_Archive.24-01-18.21-55-39.har.zip)

image

isdnfan:
I'm sorry @julien-nc I still exactly the same issue.. still `/logout?requesttoken..` followed by `/login?redirect_url=/logout?requesttoken..` I don't see any changes with the last commit. I have now idea if something is wrong with my system - can you advice please?

juliushaertl:
@julien-nc I think the code change makes sense, but with the config @isdnfan describes I see an issue as there is no singleLogoutService URL defined, so we cannot logout at the IdP and therefore are logged in back right away. However without that I also don't see a good solution except having another intermediate login step with a button that the user needs to click (or do not perform the automatic redirect to the provider when auto_logout is set.

isdnfan:
hi @juliushaertl  thank you for your response.. there is no intention to logout the user from IdP when NC session ends. the idea is exactly as you describe - after session ends on NC side the user immediately login again using IdP.. 
the problem is new login remembers previous logout URL and immediately proceed to the logout URL.. I'm wondering there is a way to catch such condition and simply replace `/login?redirect_url=/logout...` with start page?

Or maybe I'm doing something wrong in general? what is the right way to move the "login governance" completely to IdP? in other words I want my IdP to keep track about user session and NC should check very frequently if the user is logged in to IdP and deny access as soon there is no valid IdP login?

PR #815, opened 2024-03-14
Support search by email

### LINKED ISSUE #814: Support search by email
It seems that the oidc backend does not support searching user by email, only by uid and displayName : https://github.com/nextcloud/user_oidc/blob/main/lib/Db/UserMapper.php#L75

I was using the saml backend who supports search by email : https://github.com/nextcloud/user_saml/blob/master/lib/UserBackend.php#L315

As result, I can't find user in admin section when I search them by email.

user_oidc version :  5.0.1

tcoupin:
fix #814

### DIFF
--- a/lib/Db/UserMapper.php
+++ b/lib/Db/UserMapper.php
@@ -75,10 +75,16 @@ class UserMapper extends QBMapper {
 	public function find(string $search, $limit = null, $offset = null): array {
 		$qb = $this->db->getQueryBuilder();
 
-		$qb->select('*')
-			->from($this->getTableName())
+		$qb->select('user_id', 'display_name')
+			->from($this->getTableName(), 'u')
+			->leftJoin('u', 'preferences', 'p', $qb->expr()->andX(
+				$qb->expr()->eq('userid', 'user_id'),
+				$qb->expr()->eq('appid', $qb->expr()->literal('settings')),
+				$qb->expr()->eq('configkey', $qb->expr()->literal('email')))
+			)
 			->where($qb->expr()->iLike('user_id', $qb->createPositionalParameter('%' . $this->db->escapeLikeParameter($search) . '%')))
 			->orWhere($qb->expr()->iLike('display_name', $qb->createPositionalParameter('%' . $this->db->escapeLikeParameter($search) . '%')))
+			->orWhere($qb->expr()->iLike('configvalue', $qb->createPositionalParameter('%' . $this->db->escapeLikeParameter($search) . '%')))
 			->orderBy($qb->func()->lower('user_id'), 'ASC')
 			->setMaxResults($limit)
 			->setFirstResult($offset);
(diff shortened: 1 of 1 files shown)


julien-nc:
Thanks for the PR!

I would be in favor of adding email search for the `find` method but I'm not sure about adding it to the `findDisplayNames` one. @tcoupin Could you elaborate about why you made the change to `findDisplayNames` too?

The `getDisplayNames` method of all user backends is called by the server user manager in `searchDisplayName()` and `searchKnownUsersByDisplayName()`. Both those methods are not supposed to match email addresses. It's explicitly said it searches **by display name**.
https://github.com/nextcloud/server/blob/master/lib/private/User/Manager.php#L325-L340

IMO the question is: What's the real purpose of `UserInterface::getDisplayNames`?
* If it is to perform a search by display name (like the user manager expects), user_oidc should not match emails
* If it is supposed to perform the same search as `UserInterface::getUsers` but format the result differently, then yes, user_oidc should match emails.

Looking at `lib/public/UserInterface.php` and `lib/private/User/Manager.php`, it is ambiguous.

@juliushaertl WDYT?

julien-nc:
@tcoupin If you can sign your commits, great. Otherwise I could sign them when rebasing later.

tcoupin:
> @tcoupin Could you elaborate about why you made the change to findDisplayNames too?

I made same as https://github.com/nextcloud/user_saml/blob/master/lib/UserBackend.php#L315

julien-nc:
@tcoupin Ok thanks. I'll dig a bit to clarify the inconsistency between what the user manager expects and the user backend interface.
Probably better to get the same users in both find methods.

julien-nc:
@blizzz Since you implemented https://github.com/nextcloud/user_saml/blob/master/lib/UserBackend.php#L315 what's you opinion on [my previous comment](https://github.com/nextcloud/user_oidc/pull/815#issuecomment-1999488172)?

In short, the question is whether a user backend's `getDisplayNames` method should search by email address. I find it strange that the user manager assumes it only searches by display name and the implementation in user_saml searches by name and email address.

blizzz:
> @blizzz Since you implemented https://github.com/nextcloud/user_saml/blob/master/lib/UserBackend.php#L315 what's you opinion on [my previous comment](https://github.com/nextcloud/user_oidc/pull/815#issuecomment-1999488172)?
> 
> In short, the question is whether a user backend's `getDisplayNames` method should search by email address. I find it strange that the user manager assumes it only searches by display name and the implementation in user_saml searches by name and email address.

When you see the comment in line 316, you realize i did the same as the local user backend is doing :see_no_evil: 

https://github.com/nextcloud/server/blob/master/lib/private/User/Database.php#L269-L299

The semantics are confusing. Despite the documentation `searchDisplayNames` actually is not supposed to search against a display names, but for them – at least that is what all the implementations are doing since eternities (LDAP is doing the same in https://github.com/nextcloud/server/blob/master/apps/user_ldap/lib/User_LDAP.php#L538-L551). 

Conclusio: we might even add the shoe size :flat_shoe: But we should adjust the API documentation to the reality.

julien-nc:
@blizzz Thanks for the feedback.

@tcoupin Your commit is still not signed :grin: . If you want me to sign it, create another PR from a branch in this repo (not in your fork). Then I'll be able to `rebase -i` and sign.
As you're part of the GitHub Nextcloud org, you can push in branches of this repo, no need for a fork.

tcoupin:
@julien-nc sorry! now it should be good

julien-nc (approved):

PR #838, opened 2024-04-17
Fix setting avatar on login

### LINKED ISSUE #835: Avatar not set when logging in
So, I'm not sure if I'm misunderstanding something, but
* My OIDC provider has the claim `avatar` with the attribute `avatar` being the URL to the avatar of the user that just logged in.
* I added the claim to the scopes in user_oidc's admin panel
* and set `Avatar mapping` to `avatar` in `Extra attributes mapping`.

In fact, `oc_accounts` shows me this:

```
[...] avatar":{"value":"https:\/\/host\/media\/user-avatars\/foobar.jpeg","scope":"v2-local","verified":"1"} [...]
```

First of all, I didn't expect the `\/` escape here, but that's not what I'm up to.
My problem is that the user still has the default avatar set after logging in with this configuration.
I would've expected that said user has an avatar supplied from the OIDC provider now.

Am I misunderstanding something? Or is this an issue?
I dug a little bit into the codebase already, so I may be able to provide a patch, but I wanted to make sure first that I'm not holding it wrong.

Nextcloud: 28.0.4
PHP: 8.2
user_oidc: 5.0.2

julien-nc:
* bump min NC version to 26 to make sure Php >= 8.0 and we get `str_starts_with`
* bump app version to 6.0.0 because we dropped compatibility with NC 25
* add support for avatar URL and base64
    * URL: content is downloaded and avatar is set on the fly
    * base64: support strings which begin with `data:image/png;base64,` and `data:image/jpeg;base64,`
* mention avatar support and details in the README

@juliushaertl What do you think about supporting URLs, in terms of security?

closes #835

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -24,9 +24,6 @@ jobs:
         include:
             - php-versions: 8.1
               databases: mysql
-              server-versions: stable25
-            - php-versions: 8.1
-              databases: mysql
               server-versions: stable26
             - php-versions: 8.1
               databases: mysql
@@ -34,6 +31,9 @@ jobs:
             - php-versions: 8.1
               databases: mysql
               server-versions: stable28
+            - php-versions: 8.1
+              databases: mysql
+              server-versions: stable29
             - php-versions: 8.1
               databases: mysql
               server-versions: master
--- a/.github/workflows/phpunit.yml
+++ b/.github/workflows/phpunit.yml
@@ -19,12 +19,14 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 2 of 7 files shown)


juliushaertl:
> What do you think about supporting URLs, in terms of security?

I'd say this is fine as long as it requires configuring the avatar mapping. In that case the source would need to be somewhat trusted by the admin, but I'd say this is something we can require the admin of the Nextcloud/IdP to consider during configuration.

konradmoesch:
Is there anything still to be done that I can help with? For me, this PR looks complete.

andrevmatos:
This would be a great improvement, also to work with Authentik's user avatar file upload which becomes base64 urls

PR #848, opened 2024-04-29
Support missing end_session_endpoint in discovery during single logout

### LINKED ISSUE #847: Errors in logs regarding undefined array key 'end_session_endpoint'
Hi,

using the following configuration
```json
{
    "system": {
[...]
        "allow_user_to_change_display_name": false,
        "lost_password_link": "disabled",
        "user_oidc": {
            "use_pkce": true
        }
    },
    "apps": {
[...]
       "user_oidc": {
            "installed_version": "5.0.2",
            "types": "authentication",
            "enabled": "yes",
            "allow_multiple_user_backends": "1",
            "provider-1-checkBearer": "0",
            "provider-1-sendIdTokenHint": "0",
            "provider-1-uniqueUid": "0",
            "provider-1-mappingUid": "preferred_username",
            "provider-1-groupProvisioning": "1",
            "provider-1-providerBasedId": "0",
            "provider-1-jwksCache": ""
            "provider-1-jwksCacheTimestamp": "1714390135"
        },
[...]
}
```
with each oidc login and logout the log is filling with `Undefined array key "end_session_endpoint" at /var/www/html/custom_apps/user_oidc/lib/Controller/LoginController.php#597`.

Our oidc provider doesn't provide an appropriate URL and the setting is marked as optional, thus showing up with `null` when I look at `occ --output=json user_oidc:provider`.

Is there a workaround or will this require a fix accepting the `null` value?

Thanks - seb

julien-nc:
closes #847 
Avoid warnings when the Idp does not provide `end_session_endpoint` in the discovery.

@sebrhex Can you try to apply this change manually and check the warning log is not produced anymore?

### DIFF
--- a/lib/Controller/LoginController.php
+++ b/lib/Controller/LoginController.php
@@ -594,7 +594,7 @@ class LoginController extends BaseOidcController {
 
 				// Check if a custom end_session_endpoint is deposited otherwise use the default one provided by the openid-configuration
 				$discoveryData = $this->discoveryService->obtainDiscovery($provider);
-				$defaultEndSessionEndpoint = $discoveryData['end_session_endpoint'];
+				$defaultEndSessionEndpoint = $discoveryData['end_session_endpoint'] ?? null;
 				$customEndSessionEndpoint = $provider->getEndSessionEndpoint();
 				$endSessionEndpoint = $customEndSessionEndpoint ?: $defaultEndSessionEndpoint;
 


sebrhex:
Thank you for the quick reaction, I'll try it later today :)

sebrhex:
Thank you, I can confirm that after modifying the code line the error doesn't occur any more.

julien-nc:
Thank you for the feedback. This will be included in the next release.

PR #860, opened 2024-05-07
Fix and refactor upsert command

### LINKED ISSUE #858: Perform "Update Provider" function via occ command
### Description
I have a script that runs on nextcloud startup to perform configuration commands via occ. I have everything setup for user_oidc. The issue I am having is if I just run the commands and try to log in my group mapping is giving me unique identifiers.
### Current Solution
All I do to resolve this issue right now is open the UI settings and simply press update provider. I am not changing any options. I understand why this is the case, I just wish there was a "Update" or "Test" configuration command that could perform this same functionality easily via OCC command.

#### Current Setup Commands:
```bash
php occ app:enable user_oidc
php occ user_oidc:provider "Example SSO" --clientid="${OPENID_CLIENT}" --clientsecret="${OPENID_SECRET}" --discoveryuri="${OPENID_URL}" --mapping-uid=oid --mapping-groups=groups
php occ config:app:set --value 1 user_oidc provider-1-groupProvisioning
php occ config:system:set user_oidc single_logout --value false --type=boolean
```

julien-nc:
* Fix unique-uid default value
* Add 'group-provisioning' option
* Refactor upsert command to avoid future mistakes
* Explicitly define default values for boolean provider settings

closes #858

### DIFF
--- a/lib/Command/UpsertProvider.php
+++ b/lib/Command/UpsertProvider.php
@@ -45,32 +45,110 @@ class UpsertProvider extends Base {
 	private $crypto;
 
 	private const EXTRA_OPTIONS = [
-		'unique-uid' => ['shortcut' => null, 'mode' => InputOption::VALUE_OPTIONAL, 'description' => 'Flag if unique user ids shall be used or not. 1 to enable (default), 0 to disable', 'default' => null, 'setting_key' => ProviderService::SETTING_UNIQUE_UID],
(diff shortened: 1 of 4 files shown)


lib/Command/UpsertProvider.php
```diff
@@ -44,6 +44,35 @@ class UpsertProvider extends Base {
 	/** @var ICrypto */
 	private $crypto;
 
+	private const EXTRA_OPTIONS = [
+		'unique-uid' => ['shortcut' => null, 'mode' => InputOption::VALUE_OPTIONAL, 'description' => 'Flag if unique user ids shall be used or not. 1 to enable (default), 0 to disable', 'default' => null, 'setting_key' => ProviderService::SETTING_UNIQUE_UID],
```
juliushaertl, line 48: Would prefer to add linebreaks to the nested array to make it a bit more readable ;)

juliushaertl (approved):
Looks good, just a readability nitpick :)

julien-nc:
Thanks for the review.
Laziness + Vim macros == happiness

PR #862, opened 2024-05-10
fix(tests): Improve the stub so it's not confusing IDEs

nickvergessen:
### Steps
1. Checkout this app
2. Checkout another app that uses commands
3. Open the source code of a command
4. See that your IDE complains about using returned `void`s

Before | After
---|---
Bildschirmfoto vom 2024-05-10 09-00-20 | Bildschirmfoto vom 2024-05-10 09-00-05

### DIFF
--- a/tests/stubs/oc_core_command_base.php
+++ b/tests/stubs/oc_core_command_base.php
@@ -24,6 +24,9 @@ namespace OC\Core\Command {
 		public function setName(string $name) {
 		}
 
+		/**
+		 * @return $this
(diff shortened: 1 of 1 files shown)


julien-nc (approved):
:+1:

nickvergessen:
one of them also exists in older stable branches, would be cool to backport it :see_no_evil:

nickvergessen:
ah, they all use main. So it will heal. cool!

PR #863, opened 2024-05-10
Add return annotation to setName in command stub

julien-nc:
PhpStorm still complained about setName, which makes sense. I wonder why it didn't for you @nickvergessen .


nickvergessen (approved):
Mostlikely because there some of the tools that bring in symfony commands win?
grafik

grafik

Don't really understand it, but well that's why I really dislike the fact we don't have a OCP command base....

PR #864, opened 2024-05-10
Add issuer, audience and azp checks in bearer token validator

### LINKED ISSUE #856: Audience claim verification check with Bearer token
### Description
I am currently in a situation where i have an `access_token` for `nextcloud` which i got from `keycloak`. The `user_oidc` app allows a setting to make API request as bearer-auth using the `access_token`. In the `access_token` i have an audience claim for example `another-client` other than `nextcloud` itself.

for example:

```json
{
    "exp": 1714648426,
    "iat": 1714647826,
    "auth_time": 1714647781,
    "jti": "8ebabf79-7a87-4c74-86fe-58e24300677b",
    "iss": "https://keycloak.local/realms/opendesk",
    `"aud": "https://openproject.local"`,
    "sub": "6b65f442-e0a2-4e2d-ac0a-32506dd57f41",
    "typ": "Bearer",
    "azp": "nextcloud",
    "session_state": "28b06352-4cb6-4fca-9487-11795e1ec7a2",
}
```
In the above payload of the `access_token` the audience claim is `"aud": "https://openproject.local"` which is not `nextcloud`. And i am able to make API request with it.

I have some question regarding it.
- Does the `user_oidc` app check and verify the audience claim in the `access_token` for API request?
- Or the `user_oidc` is missing to check the audience claim?
- Or its intended feature?
- Or any other reasons why the audience claim is not checked and verified when making API request?

julien-nc:
The bearer token validation is less complete than the login controller one.

* add issuer, audience and azp (authorized party) checks in bearer token validator
* make it possible to disable bearer token audience check via config.php.

closes #856

### DIFF
--- a/lib/User/Validator/SelfEncodedValidator.php
+++ b/lib/User/Validator/SelfEncodedValidator.php
@@ -99,18 +99,11 @@ class SelfEncodedValidator implements IBearerTokenValidator {
 				return null;
 			}
 
-			// If the ID Token contains multiple audiences, the Client SHOULD verify that an azp Claim is present.
-			// If an azp (authorized party) Claim is present, the Client SHOULD verify that its client_id is the Claim Value.
-			if (is_array($tokenAudience) && count($tokenAudience) > 1) {
-				if (isset($payload->azp)) {
-					if ($payload->azp !== $providerClientId) {
-						$this->logger->debug('This token is not for us, authorized party (azp) is different than the client ID');
-						return null;
-					}
-				} else {
-					$this->logger->debug('Multiple audiences but no authorized party (azp) in the id token');
-					return null;
-				}
+			// ref https://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation
+			// If the azp claim is present, it should be the client ID
+			if (isset($payload->azp) && $payload->azp !== $providerClientId) {
+				$this->logger->debug('This token is not for us, authorized party (azp) is different than the client ID');
+				return null;
 			}
 		}
 
--- a/lib/User/Validator/SelfEncodedValidator.php
+++ b/lib/User/Validator/SelfEncodedValidator.php
@@ -31,6 +31,7 @@ use OCA\UserOIDC\Service\ProviderService;
 use OCA\UserOIDC\User\Provisioning\SelfEncodedTokenProvisioning;
 use OCA\UserOIDC\Vendor\Firebase\JWT\JWT;
 use OCP\AppFramework\Utility\ITimeFactory;
+use OCP\IConfig;
 use Psr\Log\LoggerInterface;
 use Throwable;
 
@@ -42,11 +43,19 @@ class SelfEncodedValidator implements IBearerTokenValidator {
 	private $logger;
 	/** @var ITimeFactory */
 	private $timeFactory;
+	/** @var IConfig */
+	private $config;
 
-	public function __construct(DiscoveryService $discoveryService, LoggerInterface $logger, ITimeFactory $timeFactory) {
+	public function __construct(
+		DiscoveryService $discoveryService,
+		LoggerInterface $logger,
+		ITimeFactory $timeFactory,
+		IConfig $config
+	) {
 		$this->discoveryService = $discoveryService;
 		$this->logger = $logger;
 		$this->timeFactory = $timeFactory;
+		$this->config = $config;
 	}
 
 	public function isValidBearerToken(Provider $provider, string $bearerToken): ?string {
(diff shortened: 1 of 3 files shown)


lib/User/Validator/SelfEncodedValidator.php
```diff
@@ -70,6 +79,36 @@ public function isValidBearerToken(Provider $provider, string $bearerToken): ?st
...
+			// If the ID Token contains multiple audiences, the Client SHOULD verify that an azp Claim is present.
+			// If an azp (authorized party) Claim is present, the Client SHOULD verify that its client_id is the Claim Value.
+			if (is_array($payload->aud) && count($payload->aud) > 1) {
+				if (isset($payload->azp)) {
+					if ($payload->azp !== $provider->getClientId()) {
+						$this->logger->debug('This token is not for us, authorized party (azp) is different than the client ID');
+						return null;
+					}
+				} else {
+					$this->logger->debug('Multiple audiences but no authorized party (azp) in the id token');
```
SagarGi, line 106:
```suggestion
					$this->logger->debug('Multiple audiences but no authorized party (azp) in the access token');
```
or simple token may be?

lib/User/Validator/SelfEncodedValidator.php
```diff
@@ -70,6 +79,36 @@ public function isValidBearerToken(Provider $provider, string $bearerToken): ?st
...
+		$oidcSystemConfig = $this->config->getSystemValue('user_oidc', []);
+		$checkAudience = !isset($oidcSystemConfig['selfencoded_bearer_validation_audience_check'])
+			|| !in_array($oidcSystemConfig['selfencoded_bearer_validation_audience_check'], [false, 'false', 0, '0'], true);
+		if ($checkAudience) {
+			if (!($payload->aud === $provider->getClientId() || in_array($provider->getClientId(), $payload->aud, true))) {
+				$this->logger->debug('This token is not for us, the audience does not match the client ID');
+				return null;
+			}
+
+			// If the ID Token contains multiple audiences, the Client SHOULD verify that an azp Claim is present.
```
SagarGi, line 97:
```suggestion
			// If the Access Token contains multiple audiences, the Client SHOULD verify that an azp Claim is present.
```

lib/User/Validator/SelfEncodedValidator.php
julien-nc: ID tokens are JWT tokens. Access tokens are more or less equivalent to OAuth tokens. This log message is correct IMO.

lib/User/Validator/SelfEncodedValidator.php
```diff
@@ -70,6 +79,36 @@ public function isValidBearerToken(Provider $provider, string $bearerToken): ?st
...
+		if ($payload->iss !== $discovery['issuer']) {
+			$this->logger->debug('This token is issued by the wrong issuer, it does not match the one from the discovery endpoint');
+			return null;
+		}
+
+		$oidcSystemConfig = $this->config->getSystemValue('user_oidc', []);
+		$checkAudience = !isset($oidcSystemConfig['selfencoded_bearer_validation_audience_check'])
+			|| !in_array($oidcSystemConfig['selfencoded_bearer_validation_audience_check'], [false, 'false', 0, '0'], true);
+		if ($checkAudience) {
+			if (!($payload->aud === $provider->getClientId() || in_array($provider->getClientId(), $payload->aud, true))) {
```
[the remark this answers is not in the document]
julien-nc: Audience can be a string or an array. The check is now safer. Any better?

julien-nc:
@wielinde Could you say again what was your concern about the `azp` check?

The current check (when validating the token for API calls) is:
If there are multiple audiences in the token (the `aud` attribute of the token is an array), we want the `azp` token attribute (authorized party) to be equal to the Oidc client ID.

This was implemented following points 4 and 5 of https://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation

Do you think this check only makes sense to authenticate in NC but not when validating a token used to make API calls?

wielinde:
@julien-nc I understand the specification that you have linked above in the following manner:

Only when there is an `azp` is given, then it shall be taken into account. However, that it is present in a token is not the default. In a vanilla OIDC setup, the `azp` shall not be added to a token, unless required by a OIDC extension.

And I am not aware of any such OIDC extension. So from my understanding, I would remove the requirement that the an `azp` must be given whenever an array of `aud` is present. And when `azp` is given, then it needs to be in the list auf `aud` AND be the Nextcloud client. Does that make sense?

// FYI @ba1ash

julien-nc:
@wielinde I can't remember where I found a different description of how the azp check should be done. Anyway let's stick to this one.
For both login token and bearer token, the check is now: If the token has the `azp` claim, it should be equal to the Oidc client ID.
If audience check is disabled for bearer token validation, the `azp` is not checked even if it's present.

All good?

wielinde:
@julien-nc From my understanding of the spec, that looks good, now. Thank you.

juliushaertl (approved):
Code looks good 👍

col-panic:
I'm using  `user_oidc` to protect my API. Now since this patch, I cannot access Nextcloud using API anymore, the output is `This token is not for us, authorized party (azp) is different than the client ID`

I found the setting `selfencoded_bearer_validation_audience_check` but it seems to catch only when coming in via `https://github.com/nextcloud/user_oidc/blob/f5bb30caef3a31ba19fe3f44da256b46cac90e9f/lib/User/Validator/SelfEncodedValidator.php#L90` not when coming in via https://github.com/nextcloud/user_oidc/blob/f5bb30caef3a31ba19fe3f44da256b46cac90e9f/lib/Controller/LoginController.php#L506

Check for example https://github.com/search?q=repo%3Anextcloud%2Fuser_oidc+%22This+token+is+not+for+us%2C+authorized+party+%28azp%29+is+different+than+the+client+ID%22&type=code

Could you please verify, that this option does also disable `azp` and `aud` checkin when used via API?

Should I create a new issue for this?

PR #868, opened 2024-05-14
Ignore keys using a P-521 curve

### LINKED ISSUE #823: DomainException: Unrecognised or unsupported EC curve at apps/user_oidc/lib/Vendor/Firebase/JWT/JWK.php line 145
With certain configurations of Keycloak as OIDC the login fails and the above mentioned error is posted to the nextcloud log.
This regression was introduced by f7b68dd which bumped the version of firebase/php-jwt which contains the buggy code.

ref: https://github.com/firebase/php-jwt/issues/561

julien-nc:
php-jwt dropped support for P-521 curve. Parsing a keyset which contains a key using this curve will now fail. We could either wait for php-jwt to do something about that. We could also remove this key from the keyset we receive from the IdP.

This is an attempt to fix #823 by removing any key using a P-521 curve from the discovered keys.

### DIFF
--- a/lib/Service/DiscoveryService.php
+++ b/lib/Service/DiscoveryService.php
@@ -156,6 +156,12 @@ class DiscoveryService {
 		}
 
 		foreach ($jwks['keys'] as $index => $key) {
+			// php-jwt fails in JWK::parseKeySet the keyset contains one key with P-521 curve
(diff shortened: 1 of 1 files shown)


promasu:
I have this code running in our installation for quite some time and everything is perfectly fine. Are we able to proceed with getting this merged to get this bug resolved?

PR #880, opened 2024-06-07
Set group displayname

julien-nc:
This is #626 with the fixed tests.

closes #626

cc @towo @jofmi @yavorivanov-cw @nanu-c


julien-nc:
@towo Please don't be sorry, I was super slow to react on your PR.

jofmi:
thanks a lot!

Boehmer86:
Thanks also from my end! Is there any indication when a new release will be done which contains this? i would like to test it with Keycloak and my NextCloud as i just encountered the same issue and likely step to resolution.

PR #884, opened 2024-06-17
 feat: restrict login to users matching a certain group

### LINKED ISSUE #867: Restrict login to users matching a certain group
We would like to limit access to Nextcloud to users having certain groups. So when a user authenticates with OAuth, it should check the groups coming from the OAuth provider before a user is created / allowed to login in. If the user is not on a whitelist (regex), the user login is rejected.

## Background

We have a member database containing many members from different divisions. Some divisions have their own Nextcloud instance, and it would be great to allow them to use the member database for login. But other divisions should not be able to use the Nextcloud instance. The user story is described in https://github.com/hitobito/hitobito_jubla/issues/74 (german)

## Proposal

- Add a new field in the settings for setting a whitelist regex for which groups (gid) are synchronized with Nextcloud
- Add a new checkbox in the settings "Restrict login for users without whitelisted groups"
- During the authentication process, check if the user is in at least one group that is part of the whitelist
- During the group synchronisation, only add and remove groups that are whitelisted. (this would also solve #866)

What do you think about this proposal?

bergerar:
Fixes  #867

### DIFF
--- a/lib/Controller/LoginController.php
+++ b/lib/Controller/LoginController.php
@@ -458,10 +458,10 @@ class LoginController extends BaseOidcController {
 		$restrictLoginToGroups = $this->providerService->getSetting($providerId, ProviderService::SETTING_RESTRICT_LOGIN_TO_GROUPS, '0');
 		if($restrictLoginToGroups === '1') {
 			$syncGroups = $this->provisioningService->getSyncGroupsOfToken($providerId, $idTokenPayload);
-			$this->logger->debug('Prevented user from login as user is not part of a whitelisted group');
 
 			if($syncGroups === null || count($syncGroups) === 0) {
-				$message = $this->l10n->t('You are not allowed to login');
+				$this->logger->debug('Prevented user from login as user is not part of a whitelisted group');
+				$message = $this->l10n->t('You do not have permission to log in to this instance. If you believe this is an error, please contact an Administrator.');
 				return $this->build403TemplateResponse($message, Http::STATUS_FORBIDDEN, ['reason' => 'user not allowed to login']);
 			}
 		}
--- a/lib/Controller/LoginController.php
+++ b/lib/Controller/LoginController.php
@@ -461,8 +461,8 @@ class LoginController extends BaseOidcController {
 
 			if($syncGroups === null || count($syncGroups) === 0) {
 				$this->logger->debug('Prevented user from login as user is not part of a whitelisted group');
-				$message = $this->l10n->t('You do not have permission to log in to this instance. If you believe this is an error, please contact an Administrator.');
-				return $this->build403TemplateResponse($message, Http::STATUS_FORBIDDEN, ['reason' => 'user not allowed to login']);
+				$message = $this->l10n->t('You do not have permission to log in to this instance. If you think this is an error, please contact an administrator.');
+				return $this->build403TemplateResponse($message, Http::STATUS_FORBIDDEN, ['reason' => 'user not in any whitelisted group']);
 			}
 		}
 
--- a/lib/Controller/LoginController.php
+++ b/lib/Controller/LoginController.php
@@ -456,10 +456,10 @@ class LoginController extends BaseOidcController {
 
 		// prevent login of users that are not in a whitelisted group (if activated)
 		$restrictLoginToGroups = $this->providerService->getSetting($providerId, ProviderService::SETTING_RESTRICT_LOGIN_TO_GROUPS, '0');
-		if($restrictLoginToGroups === '1') {
+		if ($restrictLoginToGroups === '1') {
 			$syncGroups = $this->provisioningService->getSyncGroupsOfToken($providerId, $idTokenPayload);
 
-			if($syncGroups === null || count($syncGroups) === 0) {
+			if ($syncGroups === null || count($syncGroups) === 0) {
 				$this->logger->debug('Prevented user from login as user is not part of a whitelisted group');
 				$message = $this->l10n->t('You do not have permission to log in to this instance. If you think this is an error, please contact an administrator.');
 				return $this->build403TemplateResponse($message, Http::STATUS_FORBIDDEN, ['reason' => 'user not in any whitelisted group']);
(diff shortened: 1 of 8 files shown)


edward-ly:
Closing as discussed in https://github.com/nextcloud/user_oidc/issues/867#issuecomment-2406113586.

Stefanqn:
I'm quite surprised about this decision: the referenced discussion describes a cumbersome workaround, including creating custom Keycloak flows and carefully adding the Keycloak plugin to multiple locations in these flows. Furthermore, it's an unofficial, single-developer plugin and it limits the solution space to Keycloak. On the other hand, a pull request with a proper solution at PEP (Policy Enforcement Point) exists and gets rejected. Would you please elaborate on this decision?

edward-ly:
We happily welcome open discussion from anyone, not just those already mentioned. We're not saying that this feature isn't useful, but rather that it's better to implement it as a separate app or use LDAP as an alternative.

I agree that the Keycloak extension isn't a perfect solution, but what about my point above? What does this PR provide that another app doesn't (user_ldap or otherwise)?

Stefanqn:
Thank you Edward for keeping an open mind on this. 

Regarding your point(s):
a)  use LDAP as an alternative: I'm quite confused about this suggestion as it would result in switching to a completely different protocol for authorization - one many don't have in place. Maybe you can describe your idea in more detail.
b)  implement it as a separate app: what would be the motivation to do so? You mentioned user_LDAP directly implementing this, so why do it differently for user_oidc?

edward-ly:
Let's continue the discussion in https://github.com/nextcloud/user_oidc/issues/867#issuecomment-2435647720 in which I made a response there.

edward-ly:
Reopening as per https://github.com/nextcloud/user_oidc/issues/867#issuecomment-2438267667, will test and merge within the next few days.

edward-ly (changes requested):
Please rebase the branch to start at the latest (main) commit and resolve any conflicts before we can start testing and merging, thank you.

bergerar:
I rebased the pull request, unit tests and lints are passing and I did some manual tests in my setup of the feature.

lib/Controller/LoginController.php
```diff
@@ -549,6 +549,18 @@ public function code(string $state = '', string $code = '', string $scope = '',
...
 		}
 
+		// prevent login of users that are not in a whitelisted group (if activated)
+		$restrictLoginToGroups = $this->providerService->getSetting($providerId, ProviderService::SETTING_RESTRICT_LOGIN_TO_GROUPS, '0');
+		if($restrictLoginToGroups === '1') {
+			$syncGroups = $this->provisioningService->getSyncGroupsOfToken($providerId, $idTokenPayload);
+
+			if($syncGroups === null || count($syncGroups) === 0) {
+				$this->logger->debug('Prevented user from login as user is not part of a whitelisted group');
+				$message = $this->l10n->t('You do not have permission to log in to this instance. If you believe this is an error, please contact an Administrator.');
```
julien-nc, line 559:
```suggestion
				$message = $this->l10n->t('You do not have permission to log in to this instance. If you think this is an error, please contact an administrator.');
```

lib/Controller/LoginController.php
```diff
@@ -549,6 +549,18 @@ public function code(string $state = '', string $code = '', string $scope = '',
...
+				return $this->build403TemplateResponse($message, Http::STATUS_FORBIDDEN, ['reason' => 'user not allowed to login']);
```
julien-nc, line 560:
```suggestion
				return $this->build403TemplateResponse($message, Http::STATUS_FORBIDDEN, ['reason' => 'user not in any whitelisted group']);
```

src/components/SettingsForm.vue
```diff
@@ -257,6 +257,21 @@
...
 			{{ t('user_oidc', 'This will create and update the users groups depending on the groups claim in the id token. The Format of the groups claim value should be [{gid: "1", displayName: "group1"}, ...] or ["group1", "group2", ...]') }}
 		</p>
+		<p>
+			<label for="group-whitelist-regex">{{ t('user_oidc', 'Group whitelist regex') }}</label>
+			<input id="group-whitelist-regex"
+				v-model="localProvider.settings.groupWhitelistRegex"
+				type="text">
+		</p>
+		<p class="settings-hint">
+			{{ t('user_oidc', 'Only groups matching the whitelist regex will be created, updated and deleted by the group claim') }}
```
julien-nc, line 267:
```suggestion
			{{ t('user_oidc', 'Only groups matching the whitelist regex will be created, updated and deleted by the group claim. For example: {regex} allows all groups which ID starts with {substr}', { regex: '/^blue/', substr: 'blue' }) }}
```
Or anything more descriptive.

src/components/SettingsForm.vue
```diff
@@ -257,6 +257,21 @@
...
+		</p>
+		<NcCheckboxRadioSwitch :checked.sync="localProvider.settings.restrictLoginToGroups" wrapper-element="div">
+			{{ t('user_oidc', 'Restrict login for users without whitelisted groups.') }}
```
julien-nc, line 270:
```suggestion
			{{ t('user_oidc', 'Restrict login for users that are not in any whitelisted group') }}
```

src/components/SettingsForm.vue
```diff
@@ -257,6 +257,21 @@
...
+		</NcCheckboxRadioSwitch>
+		<p class="settings-hint">
+			{{ t('user_oidc', 'Users that are not part of a whitelisted group are not created and can not login') }}
```
julien-nc, line 273:
```suggestion
			{{ t('user_oidc', 'Users that are not part of any whitelisted group are not created and can not login') }}
```

julien-nc (changes requested):
Looks sane to me. Tested, works fine :heavy_check_mark: 

A small improvement could be to check if the leading and trailing slashes are there and add them if not.

The labels/hints in the settings need to be as much descriptive as possible. See inline comments.

bergerar:
@julien-nc  the delimiters are now added automatically if the first one is missing.

There was one check failing, but it was the same one as on the main branch (Integration tests / php8.1-mysql-stable27)

bergerar:
still had an old version of the php-cs-fixer.

I did now rebase, update install the updated dependencies and run `composer run cs:fix`.

@julien-nc is should now pass all tests

julien-nc (approved):

bergerar:
Thanks for merging and your work in general on open source. @julien-nc

julien-nc:
@bergerar Thank you for your patience, your perseverance, your nice words and your contribution!

PR #897, opened 2024-07-17
Fix #349 Support for client_secret_basic only token_endpoint_auth_method

### LINKED ISSUE #349: Module ignores discovered client authentication method, and always uses post
I am working with an openid connect server that only allows client_secret_basic, which is correctly discovered in the openid discovery endpoint. This module appears to ignore that setting, and always uses client_secret_post.

This module should correctly respect the discovered settings.

xataxxx:
The OIDC provider I'm using only supports client_secret_basic in their token_endpoint_auth_methods_supported.
This is coded in a way that it only attempts to use the header based auth when ONLY client_secret_basic is provided in discovery "token_endpoint_auth_methods_supported".

### DIFF
--- a/lib/Controller/LoginController.php
+++ b/lib/Controller/LoginController.php
@@ -415,7 +415,7 @@ class LoginController extends BaseOidcController {
 
 			$headers = [];
 			$tokenEndpointAuthMethod = 'client_secret_basic';
-			// Use POST only if client_secret_basic is not available
+			// Use POST only if client_secret_basic is not available as supported by the endpoint
 			if (array_key_exists('token_endpoint_auth_methods_supported', $discovery) &&
 				is_array($discovery['token_endpoint_auth_methods_supported']) &&
 				in_array('client_secret_post', $discovery['token_endpoint_auth_methods_supported']) &&
@@ -424,9 +424,8 @@ class LoginController extends BaseOidcController {
 			}
 
 			if ($tokenEndpointAuthMethod == 'client_secret_basic') {
-				$credentials = base64_encode($provider->getClientId() . ':' . $providerClientSecret);
 				$headers = [
-					'Authorization' => 'Basic ' . $credentials,
+					'Authorization' => 'Basic ' . base64_encode($provider->getClientId() . ':' . $providerClientSecret),
 					'Content-Type' => 'application/x-www-form-urlencoded',
 				];
 			} else {
--- a/lib/Controller/LoginController.php
+++ b/lib/Controller/LoginController.php
@@ -414,10 +414,33 @@ class LoginController extends BaseOidcController {
 			if ($isPkceEnabled) {
 				$requestBody['code_verifier'] = $this->session->get(self::CODE_VERIFIER); // Set for the PKCE flow
 			}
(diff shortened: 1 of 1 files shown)


julien-nc:
Thanks a lot for this PR. Much needed indeed.

According to https://openid.net/specs/openid-connect-core-1_0.html#ClientAuthentication , `client_secret_basic` should even be the default when no supported auth method is listed in the discovery.
`token_endpoint_auth_methods_supported` is described in https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata

Could you split the code into 2 sections?
* first determine which auth method should be used (default to basic auth, use body params if `token_endpoint_auth_methods_supported` is there and contains `client_secret_post`)
* then set the necessary headers or body params

Instead of always setting the body params and then unsetting them if basic auth should be used, you can set them only if basic auth is not used (remove them from the initial assignment to `$requestBody`).

Wdyt?

xataxxx:
I agree with splitting the code into picking the right auth method and then applying the correct params and headers.
I'll try to tidy up that part a bit by the end of the week.

I chose the current logic because I wanted to code to continue working like it is currently (and keep my code change minimal).

The current code always sends POST request.
If we change the default from POST to basic if no 'token_endpoint_auth_methods_supported' is specified then it could become a breaking change for some unlucky users.

So in that regard I'm not 100% sure changing the default behaviour is the right thing to do (for Nextcloud users), even though it would be the correct thing by the spec.

If it was up to me, I would change the default to basic like the spec says and hope for the best :)

julien-nc:
I think it's pretty safe to assume that Oidc providers which don't declare a `token_endpoint_auth_methods_supported` will support basic auth as it's the recommended default in the official Oidc specs.

Let's follow the official specs in user_oidc.

xataxxx:
Changed the code, now defaulting to client_secret_basic if no supported methods is specified AND even if both basic and post is specified (as basic is default, I gave it preference).

julien-nc:
Thanks a lot. Can you sign all the commits? There are instructions in the logs of the failing CI action. If you can't I'll force merge :grin:

xataxxx:
I think I got it.

julien-nc (approved):

hendrik1120:
Hi guys, this unfortunately was a breaking change as you have effectively changed the default `token_endpoint_auth_method`.

Just because the provider generally does support the `token_endpoint_auth_method` it is not guaranteed that the client is also allowed to use it.
In a specific case, Authelia was configured to only allow `client_secret_post` for the nextcloud client, as this was the only supported `auth_method` at the time. Just because Authelia in this case is able to support all auth methods defined by the spec, it's not guaranteed the client can also use all of them. This is in compliance with the spec as an oidc client generally should only have one `auth_method` defined.

The specification is not entirely clear around this area, but I would argue that the `token_endpoint_auth_methods_supported` together with the`token_endpoint_auth_method` values are intended for the dynamic client registration section of the spec. As this feature is not available in this plugin, a configuration option to change the default `token_endpoint_auth_method` might have been the best solution instead of introducing this dynamic behavior.

PR #921, opened 2024-08-20
Make audience and azp checks optional 

### LINKED ISSUE #918: `azp` and `selfencoded_bearer_validation_audience_check` configuration
Following up to https://github.com/nextcloud/user_oidc/issues/856#issuecomment-2290757195 

1. `selfencoded_bearer_validation_audience_check` can only be deactivated by directly configuring this in `config.php` - a call with `occ config:app:set user_oidc selfencoded_bearer_validation_audience_check --value=0` does not have any effect. This might be of interest or not - I recommend documenting it on the front page if it is intended.
2. `azp` on Keycloak can not be removed from both an access-token and an ID token. So there is no way for me to skip verification of `azp` without deactivating all of the audience check. This is imho FALSE. It could be right, that `azp` is e.g. `some-client` while the `aud` includes `nextcloud` -> This is the case if I gather a ticket in the name of `some-client` but deliberately add `nextcloud` aus `aud` in the process of creating this ticket. Thus the verification of `azp` should be deactivateble without deactiving both `aud` and `azp` verification.
3. For docker usage deactivating this should be doable by means of an environment variable. I don't know if this is possible at the moment.

julien-nc:
Make audience and azp checks optional when logging in or validating a bearer token.

refs #856
closes #918

@col-panic Until the azp check is clarified, I think it's enough to be able to disable aud and azp checks independently during login or NC API requests bearer token validation. Wdyt?

### DIFF
--- a/lib/Controller/LoginController.php
+++ b/lib/Controller/LoginController.php
@@ -489,23 +489,31 @@ class LoginController extends BaseOidcController {
 		}
 
 		// Verify audience
-		$tokenAudience = $idTokenPayload->aud;
-		$providerClientId = $provider->getClientId();
-		if (
-			(is_string($tokenAudience) && $tokenAudience !== $providerClientId)
+		$checkAudience = !isset($oidcSystemConfig['login_validation_audience_check'])
+			|| !in_array($oidcSystemConfig['login_validation_audience_check'], [false, 'false', 0, '0'], true);
+		if ($checkAudience) {
+			$tokenAudience = $idTokenPayload->aud;
+			$providerClientId = $provider->getClientId();
+			if (
+				(is_string($tokenAudience) && $tokenAudience !== $providerClientId)
 				|| (is_array($tokenAudience) && !in_array($providerClientId, $tokenAudience, true))
-		) {
-			$this->logger->debug('This token is not for us');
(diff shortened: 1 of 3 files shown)


col-panic:
@julien-nc I think that its good, offers the most flexibility. Do you know why they have to be hardcoded in config.php and can't be set with `occ`?

lib/Controller/LoginController.php
```diff
@@ -489,23 +489,31 @@ public function code(string $state = '', string $code = '', string $scope = '',
 		}
 
 		// Verify audience
-		$tokenAudience = $idTokenPayload->aud;
-		$providerClientId = $provider->getClientId();
-		if (
-			(is_string($tokenAudience) && $tokenAudience !== $providerClientId)
+		$checkAudience = !isset($oidcSystemConfig['login_validation_audience_check'])
+			|| !in_array($oidcSystemConfig['login_validation_audience_check'], [false, 'false', 0, '0'], true);
```
juliushaertl, line 493: I'd say it is also fine to just support actual boolean values, but fine with me 😁

juliushaertl (approved):

col-panic:
@juliushaertl wdyt when will this be integrated? Any idea of the timing for the next release?

PR #946, opened 2024-09-19
fix(LoginController): revert default `token_endpoint_auth_method` value

edward-ly:
Fixes a breaking change first reported in https://github.com/nextcloud/user_oidc/pull/897#issuecomment-2304308634.

### DIFF
--- a/lib/Controller/LoginController.php
+++ b/lib/Controller/LoginController.php
@@ -414,13 +414,13 @@ class LoginController extends BaseOidcController {
 			}
 
 			$headers = [];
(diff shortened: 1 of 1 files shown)


julien-nc (approved):
:+1: Nice!

tblaschke:
Not to be ungrateful, but this effectively created a breaking change AGAIN. This time even within a minor version. Due to the last change every authelia config had to be changed to accept ‘client_secret_basic’ and now it has to be reverted to accept ‘client_secret_post’. Please handle the breaking chaining with a bit more caution and introduce them with major versions.

PR #954, opened 2024-09-26
New system config flag to disable user creation in soft auto provisioning

julien-nc:
We are missing a use case in soft auto provisioning. Some people might want to only accept Oidc logins from users that already exist in other backends (like when auto provisioning is disabled) but still want user_oidc to set the mapped attributes to the user (like when soft auto provisioning is enabled and the user already exists in another backend).

There were 2 ways to do that:
1. Add an option that makes it happen when auto provisioning is disabled (like `force_attribute_provisioning`)
2. Add an option that forbids account creation when soft auto provisioning is enabled

We chose 2. with @come-nc thinking it's slightly easier to grasp when configuring user_oidc.


lib/Service/ProvisioningService.php
```diff
@@ -149,6 +109,17 @@ public function provisionUser(string $tokenUserId, int $providerId, object $idTo
 		if ($existingLocalUser !== null) {
 			$user = $existingLocalUser;
 		} else {
+			// if disable_account_creation is true, user_oidc should not manage user
+			// - no user creation
+			// - do not use the existing users managed by the user_oidc backend
```
come-nc, line 114: Hum but what if user_oidc created some users and then I enable the new option, these users disappear? Or at least cannot login anymore?

julien-nc: They can't login... I took the easy way. I can change that behaviour. Do you think it makes more sense to let them in?

come-nc (approved):

PR #955, opened 2024-10-01
Backchannel logout endpoint should only return 200 or 400

julien-nc:
According to https://openid.net/specs/openid-connect-backchannel-1_0.html#rfc.section.2.6 , the backchannel logout should only respond with 200 or 400 status code responses. With bruteforce protection, 429 can be returned.

Should we add a rate limit instead? What would be reasonable values?

### DIFF
--- a/lib/Controller/LoginController.php
+++ b/lib/Controller/LoginController.php
@@ -671,7 +671,6 @@ class LoginController extends BaseOidcController {
 	 *
 	 * @PublicPage
 	 * @NoCSRFRequired
-	 * @BruteForceProtection(action=userOidcBackchannelLogout)
 	 *
 	 * @param string $providerIdentifier
 	 * @param string $logout_token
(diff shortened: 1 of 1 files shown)


edward-ly:
I'm not sure if either protection makes sense on the logout operation. I could be wrong, but I don't see what kind of vulnerability could be exploited from this endpoint alone.

julien-nc:
As the backchannel logout endpoint is public, one can bruteforce it to logout people. But the logout token is pretty long so the chances such attack succeeds are very low. Even if it succeeds, it's just logging out someone.

I thought this endpoint could be the target of a DoS attack because it makes NC do a network request to the Idp (to the discovery endpoint) but this is cached.

@edward-ly I agree, it seems ok to not put any protection there.

Waiting for the :heavy_check_mark: of @juliushaertl to merge.

juliushaertl (approved):

PR #956, opened 2024-10-06
fix: re-enable PKCE by default

### LINKED ISSUE #807: enable PKCE by default
with #740 PKCE was disabled by default. According to different sources PKCE is more secure and recommended for all sorts of clients:

https://datatracker.ietf.org/doc/html/draft-ietf-oauth-security-topics-16#section-2.1.1
> Clients MUST prevent injection (replay) of authorization codes into the authorization response by attackers. Public clients MUST use PKCE [RFC7636] to this end. For confidential clients, the use of PKCE [RFC7636] is RECOMMENDED.

in general IdP MUST advertise PKCE support and user_oidc should dynamically adopt rather hard-coding one or another variant:

https://datatracker.ietf.org/doc/html/draft-ietf-oauth-security-topics-16#section-2.1.1
> Authorization servers MUST provide a way to detect their support for PKCE. To this end, they MUST either (a) publish the element code_challenge_methods_supported in their AS metadata ([[RFC8414](https://www.rfc-editor.org/info/rfc8414)]) containing the supported PKCE challenge methods (which can be used by the client to detect PKCE support) or (b) provide a deployment-specific way to ensure or determine PKCE support by the AS.

Please reverse use_pkce setting and enable PKCE by default:

```
'user_oidc' => [
    'use_pkce' => true,
],
```
in order 
- to keep backward compatibility
- use RECOMMENDED config by default
- people could still disable it if required by broken IdP

edward-ly:
Resolves #807.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,10 +88,10 @@ sudo -u www-data php var/www/nextcloud/occ config:app:set --value=0 user_oidc al
 
 This app supports PKCE (Proof Key for Code Exchange).
 https://datatracker.ietf.org/doc/html/rfc7636
-It is disabled by default and can be enabled in `config.php`:
+It is enabled by default, but can be disabled in `config.php`:
 ``` php
 'user_oidc' => [
-    'use_pkce' => true,
+    'use_pkce' => false,
 ],
 ```
 
--- a/README.md
+++ b/README.md
@@ -88,7 +88,9 @@ sudo -u www-data php var/www/nextcloud/occ config:app:set --value=0 user_oidc al
 
 This app supports PKCE (Proof Key for Code Exchange).
 https://datatracker.ietf.org/doc/html/rfc7636
-It is enabled by default, but can be disabled in `config.php`:
+Unless PKCE is not supported by the configured OpenID Connect provider,
+it is enabled by default.
+You can also manually disable it in `config.php`:
 ``` php
 'user_oidc' => [
     'use_pkce' => false,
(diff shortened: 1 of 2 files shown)


edward-ly:
Hmm, all of the integration tests on the NC master branch are failing. I can't really tell from the logs where exactly the problem is coming from, though.

julien-nc:
I fixed the CI in another PR that was not merged yet. It's now fixed. I rebased your branch on main and pushed again. Should be fine now.

julien-nc (changes requested):
Since it's now enabled by default, we could check if the provider supports our code challenge method (for safety). This information can be found in the discovery. `S256` (the method we use) should be included in the `code_challenge_methods_supported` array in the provider's discovery.
Here is what's inside a classic Keycloak's discovery:
``` json
"code_challenge_methods_supported": [
  "plain",
  "S256"
],
```

So I think we should use PKCE only if it is enabled in NC's config AND the provider supports S256.

Wdyt?

julien-nc (commented):
Nice!
Can you rebase your branch on main? Since there are no conflicts, it makes the history cleaner and easier to read.

edward-ly:
> Nice! Can you rebase your branch on main? Since there are no conflicts, it makes the history cleaner and easier to read.

Ah, good idea, this should do it, I think.

julien-nc (approved):
:+1:

PR #959, opened 2024-10-10
chore: migrate to using HTTP attributes

edward-ly:


lib/Controller/ApiController.php
```diff
@@ -58,15 +60,15 @@ public function __construct(
...
-	 *
 	 * @param int $providerId
 	 * @param string $userId
 	 * @param string|null $displayName
 	 * @param string|null $email
 	 * @param string|null $quota
 	 * @return DataResponse
 	 */
+	#[NoCSRFRequired]
+	#[FrontpageRoute(verb: 'POST', url: '/user')]
```
julien-nc, line 71: NoCSRFRequired is only there since 27 and FrontpageRoute since 29. This app supports NC 26. We can switch to method attributes when we drop support for NC 26.

Issue #968, opened 2024-10-22
Undefined variable `$providerClientId`


artonge:
When `selfencoded_bearer_validation_audience_check` is set to `false`, then [`$providerClientId` is never set](https://github.com/nextcloud/user_oidc/blob/v6.1.0/lib/User/Validator/SelfEncodedValidator.php#L93), and [the comparison fails here](https://github.com/nextcloud/user_oidc/blob/v6.1.0/lib/User/Validator/SelfEncodedValidator.php#L108)

https://github.com/nextcloud/user_oidc/blob/v6.1.0/lib/User/Validator/SelfEncodedValidator.php#L90-110

Introduced by https://github.com/nextcloud/user_oidc/issues/856

PR #969, opened 2024-10-22
fix: Ensure $providerClientId is declare

artonge:
Fix https://github.com/nextcloud/user_oidc/issues/968


julien-nc (approved):
:see_no_evil: :blue_heart:

PR #971, opened 2024-10-28
Fixes Access forbidden - State token missing

joselameira:
This commit error Access forbidden - State token missing while trying to login using Nextcloud Desktop.

Commit 99f7b730e7911459de22898e431cfcf525f0393a ignored the query part of the redirectUrl, needed for Nextcloud Desktop to succeed authentication.


julien-nc (approved):
:+1:  Thanks a lot for the fix.

julien-nc:
Could you add a space between `fn` and `($k)` so the code style check passes? I'll merge right after.

joselameira:
Hi @julien-nc

Space added!

PR #974, opened 2024-10-30
Implement token exchange mechanism for other apps

julien-nc:
Some integration apps (like integration_openproject) might need to make backend-to-backend requests to other services that use the same IdP as Nextcloud. If those services accept Oidc access tokens as a Bearer tokens for authentication, they might only accept the ones that were delivered for them (for their audience).

Token exchange is a mechanism to exchange a token that was delivered for an audience (for a specific Oidc client) against another token for another audience.

The idea here is to keep the token exchange logic in user_oidc so it saves the effort to reimplement it in all the apps that need it.

In short:
* The login token is stored in the Php session
* It is refreshed when needed
* If it can't be refreshed, we try to reauthenticate (which might be transparent for the user if the IdP session is still alive)
* We listen to the `ExchangedTokenRequestedEvent`
* We exchange the login token when receiving the event and place the result in the event

cc @individual-it @SagarGi

### DIFF
--- a/lib/Service/TokenService.php
+++ b/lib/Service/TokenService.php
@@ -1,0 +1,293 @@
+<?php
+/**
+ * @copyright Copyright (c) 2024 Julien Veyssier <julien-nc@posteo.net>
+ *
+ * @author Julien Veyssier <julien-nc@posteo.net>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+declare(strict_types=1);
+
+namespace OCA\UserOIDC\Service;
+
+use OCA\UserOIDC\AppInfo\Application;
+use OCA\UserOIDC\Db\ProviderMapper;
+use OCA\UserOIDC\Exception\TokenExchangeFailedException;
+use OCA\UserOIDC\Model\Token;
+use OCA\UserOIDC\Vendor\Firebase\JWT\JWT;
+use OCP\AppFramework\Db\DoesNotExistException;
+use OCP\AppFramework\Db\MultipleObjectsReturnedException;
+use OCP\Http\Client\IClient;
+use OCP\Http\Client\IClientService;
+use OCP\IConfig;
+use OCP\IRequest;
+use OCP\ISession;
+use OCP\IURLGenerator;
+use OCP\IUserSession;
+use OCP\PreConditionNotMetException;
+use OCP\Security\ICrypto;
+use Psr\Log\LoggerInterface;
+
+/**
+ * Token management service
+ * This is helpful to debug:
+ * tail -f data/nextcloud.log | grep "\[Token" | jq ".time,.message"
+ */
+class TokenService {
+
+	private const SESSION_TOKEN_KEY = Application::APP_ID . '-user-token';
+
+	private IClient $client;
+
+	public function __construct(
+		IClientService $clientService,
+		private ISession $session,
+		private IUserSession $userSession,
+		private IConfig $config,
+		private LoggerInterface $logger,
+		private ICrypto $crypto,
+		private IRequest $request,
+		private IURLGenerator $urlGenerator,
+		private DiscoveryService $discoveryService,
+		private ProviderMapper $providerMapper,
+	) {
+		$this->client = $clientService->newClient();
+	}
+
+	public function storeToken(array $tokenData): Token {
+		$token = new Token($tokenData);
+		$this->session->set(self::SESSION_TOKEN_KEY, json_encode($token, JSON_THROW_ON_ERROR));
+		$this->logger->debug('[TokenService] Store token');
+		return $token;
+	}
+
+	/**
+	 * Get the token stored in the session
+	 * If it has expired: try to refresh it
+	 *
+	 * @param bool $refreshIfExpired
+	 * @return Token|null Return a token only if it is valid or has been successfully refreshed
+	 * @throws \JsonException
+	 */
+	public function getToken(bool $refreshIfExpired = true): ?Token {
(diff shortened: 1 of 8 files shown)


SagarGi:
hi @julien-nc thanks for the implementation. i will checkout this PR.

SagarGi (commented):
Hi @julien-nc I manually checked the PR.
I have a question for this implementation.

As far as i have checked it you are implementing as, when ever we need an exchanged token we need to emit the event?
And if we do so then i think every time we require an exchanged token we always generate a new one?

lib/Service/TokenService.php
```diff
@@ -0,0 +1,293 @@
...
+				'audience' => $targetAudience,
+			]);
+			$body = $result->getBody();
+			$bodyArray = json_decode(trim($body), true, 512, JSON_THROW_ON_ERROR);
+			$this->logger->debug('[TokenService] Token exchange success: "' . trim($body) . '"');
+			$tokenData = array_merge(
+				$bodyArray,
+				['provider_id' => $loginToken->getProviderId()],
+			);
+			return new Token($tokenData);
```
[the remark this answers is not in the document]
julien-nc: Hmmm in my tests the exchanged token has the same "shape" as the login token, it contains an id token and an access token. So it might depend on the client config on the IdP side. We can make it so the Token object does not require an id_token.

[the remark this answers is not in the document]
julien-nc: The id_token might be used for anything, not just the login. It contains user info indeed.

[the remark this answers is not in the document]
julien-nc: Can you try to change line 41 of `lib/Model/Token.php` to:
`$this->idToken = $tokenData['id_token'] ?? null;`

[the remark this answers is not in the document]
SagarGi: > Hmmm in my tests the exchanged token has the same "shape" as the login token, it contains an id token and an access token. So it might depend on the client config on the IdP side. We can make it so the Token object does not require an id_token.
Sure, I think we do not require the id token to be stored for client to client communication for oidc based authentication.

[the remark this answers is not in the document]
julien-nc: and 32 to:
`private ?string $idToken;`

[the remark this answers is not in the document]
SagarGi: yes i did and its working!

julien-nc:
> As far as i have checked it you are implementing as, when ever we need an exchanged token we need to emit the event?
And if we do so then i think every time we require an exchanged token we always generate a new one?

Yes and yes

SagarGi:
@julien-nc We need token for each request i guess does that mean we emit that event and get token every time we make request to targeted client. would not be it too much?
CC @individual-it

individual-it:
> @julien-nc We need token for each request i guess does that mean we emit that event and get token every time we make request to targeted client. would not be it too much? CC @individual-it

yes I would do it exactly like that. The integration app emits the event every time it needs the token, that way only the oidc app has to care about expiration time, renewals, caching of tokens etc.

DaphneMuller:
@SagarGi @juliusknorr any chance this can be reviewed soon?

lib/Service/TokenService.php
```diff
@@ -0,0 +1,276 @@
...
+		} catch (\Exception|\Throwable $e) {
```
nabim777, line 271: Currently, if the audience is invalid, the code throws an error:
`Failed to exchange token, error in the exchange request.`

It would be better if it throws a more specific error for an invalid audience, as this would make it clearer and more helpful, especially for displaying in the UI.

On API, when the audience is invalid, we get status code `400` and the response:

```json
{
  "error": "invalid_client",
  "error_description": "Audience not found"
}
```

lib/Service/TokenService.php
```diff
@@ -0,0 +1,276 @@
...
+	 * If it has expired: try to refresh it
+	 *
+	 * @param bool $refreshIfExpired
+	 * @return Token|null Return a token only if it is valid or has been successfully refreshed
+	 * @throws \JsonException
+	 */
+	public function getToken(bool $refreshIfExpired = true): ?Token {
+		$sessionData = $this->session->get(self::SESSION_TOKEN_KEY);
+		if (!$sessionData) {
+			$this->logger->debug('[TokenService] getToken: no session data');
```
juliusknorr, line 73: Should this be logged at a higher level?

lib/Service/TokenService.php
```diff
@@ -0,0 +1,276 @@
...
+	 * @throws \JsonException
+	 * @throws PreConditionNotMetException
+	 */
+	public function checkLoginToken(): void {
+		$currentUser = $this->userSession->getUser();
+		if (!$this->userSession->isLoggedIn() || $currentUser === null) {
+			$this->logger->debug('[TokenService] checkLoginToken: user not logged in');
+			return;
+		}
+		if ($this->config->getUserValue($currentUser->getUID(), Application::APP_ID, 'had_token_once', '0') !== '1') {
```
juliusknorr, line 108: I think I miss understanding on why we need this extra user value check, maybe you can elaborate on that?

juliusknorr (approved):
Two minor comments, otherwise that code and general structure looks good.

Actually this is nice as it could simplify logic in https://github.com/nextmcloud/nmc_spica/blob/main/lib/Service/TokenService.php where a similar thing has been done

lib/Service/TokenService.php
julien-nc: Maybe it will flood the logs since this one is produced not only when there's a token exchange request but also when checking the login token in Application.php. If the current user is not authenticated via user_oidc, there will never be any session data.

lib/Service/TokenService.php
julien-nc: If the user never had any login token before, it means it's not using user_oidc to authenticate so we know we can skip the login token check.

Do you mean that it is enough to just do nothing if there is no login token in the session?
We need to know if there is no login token available
* because user_oidc was not used to log in (we can silently skip the check)
* OR because the session has died/disappeared (then we need to reauthenticate)

Does that make sense?

lib/Service/TokenService.php
julien-nc: Good point. The `TokenExchangeFailedException` class now has 2 additional methods:
```
public function getError(): ?string;
public function getErrorDescription(): ?string;
```
They both return a string only if the exchange request fails and returns a JSON object with the `error` and `error_description` keys.

lib/Service/TokenService.php
juliusknorr: Ah ok, makes sense then 👍

lib/Service/TokenService.php
juliusknorr: All right then debug is also fine.

nabim777:
I have reviewed on behalf of @SagarGi . After I checking manually the PR, it meets the expectations and requirements. LGTM :+1:

PR #990, opened 2024-11-20
Use absolute paths instead of relative paths

### LINKED ISSUE #989: Does not correctly generate redirection urls if Nextcloud is in a path other than /
### How to use GitHub

---

### Steps to reproduce

We are migrating from `user_saml` to `user_oidc` and we noticed that `user_oidc` generates bad redirect urls when Nextcloud is set to a path other than “/” which is done by following these steps:

https://docs.nextcloud.com/server/21/admin_manual/configuration_server/reverse_proxy_configuration.html?highlight=overwrite#overwrite-parameters

If we look at `user_saml` it has solved it by generating the absolute urls which is why it works for us:

https://github.com/nextcloud/user_saml/blob/72a8539a5447a73abf941b4b4c9e1a5cfc70fc78/appinfo/app.php#L146

In `user_oidc` does not work because it does not:

https://github.com/nextcloud/user_oidc/blob/d161cdd8b254ca90246b362165003ae2854cec1e/lib/AppInfo/Application.php#L86

https://github.com/nextcloud/user_oidc/blob/d161cdd8b254ca90246b362165003ae2854cec1e/lib/AppInfo/Application.php#L100

To fix it we would have to convert the urls to absolute urls.

### Expected behaviour

It should redirect correctly even if it is set to a path other than “/”.

### Actual behaviour

It gives a 404 because it tries to resolve urls without taking into account that it is in a path different from “/”.

### Server configuration

**Web server:** Apache/Nginx

**Database:** MySQL/Maria/SQLite/PostgreSQL

**PHP version:** 8.1/8.2/8.3

**Nextcloud version:** (see Nextcloud admin page)

<details>
<summary>List of activated apps</summary>

```
If you have access to your command line run e.g.:
sudo -u www-data php occ app:list
from within your Nextcloud installation folder
```
</details>

<details>
<summary>Nextcloud configuration</summary>

```
If you have access to your command line run e.g.:
sudo -u www-data php occ config:list system
from within your Nextcloud installation folder
```
</details>

### Browser

**Browser name:** Firefox/Chrome/Safari/…

**Browser version:** 124/125/…

**Operating system:** Windows/Ubuntu/Mac/…

<details>
<summary>Browser log</summary>

```
Insert your browser log here, this could for example include:
a) The javascript console log
b) The network log
c) ...
```

</details>

bdovaz:
Fixes #989

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -83,7 +83,7 @@ class Application extends App implements IBootstrap {
 		if ($isDefaultLogin && !$settings->getAllowMultipleUserBackEnds() && count($providers) === 1) {
 			$targetUrl = $urlGenerator->linkToRoute(self::APP_ID . '.login.login', [
 				'providerId' => $providers[0]->getId(),
-				'redirectUrl' => $redirectUrl
+				'redirectUrl' => $urlGenerator->getAbsoluteURL($redirectUrl)
 			]);
 			header('Location: ' . $targetUrl);
 			exit();
@@ -97,7 +97,7 @@ class Application extends App implements IBootstrap {
 			// FIXME: Move to IAlternativeLogin but requires boot due to db connection
 			OC_App::registerLogIn([
 				'name' => $l10n->t('Login with %1s', [$provider->getIdentifier()]),
-				'href' => $urlGenerator->linkToRoute(self::APP_ID . '.login.login', ['providerId' => $provider->getId(), 'redirectUrl' => $redirectUrl]),
+				'href' => $urlGenerator->linkToRoute(self::APP_ID . '.login.login', ['providerId' => $provider->getId(), 'redirectUrl' => $urlGenerator->getAbsoluteURL($redirectUrl)]),
 			]);
 		}
 
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -72,6 +72,7 @@ class Application extends App implements IBootstrap {
 	private function registerRedirect(IRequest $request, IURLGenerator $urlGenerator, SettingsService $settings, ProviderMapper $providerMapper): void {
 		$providers = $this->getCachedProviders($providerMapper);
 		$redirectUrl = $request->getParam('redirect_url');
+		$absoluteRedirectUrl = !empty($redirectUrl) ? $urlGenerator->getAbsoluteURL($redirectUrl) : $redirectUrl;
 
 		// Handle immediate redirect to the oidc provider if just one is configured and no other backends are allowed
 		$isDefaultLogin = false;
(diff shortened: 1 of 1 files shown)


julien-nc:
Thanks a lot. I'll review that soon.

julien-nc (changes requested):
So if I understand correctly, the issue that is fixed here is when NC is served under a sub path AND there is a reverse proxy so NC is not aware of its path and the path is overwritten with the overwrite* params in config.php. Right?

I tried with a basic setup with NC accessed at http://localhost/nc30 and when trying to reach
http://localhost/nc30/index.php/login?redirect_url=/nc30/index.php/apps/activity/all
while not being authenticated, I'm redirected to the login page
http://localhost/nc30/index.php/login?redirect_url=/nc30/index.php/apps/activity/all
and the "Login with PROVIDER" button from user_oidc points to:
http://localhost/nc30/index.php/apps/user_oidc/login/1?redirectUrl=/nc30/index.php/apps/activity/all

which is correct. No issue there.

With a reverse proxy, why isn't the issue solved by setting the overwrite* params in config.php?

Anyway, if we take this fix, I have 2 change requests:
1. Process/generate the redirect URL outside the loop (line 96) so it's not done for each provider
2. The url generation could look like this:
``` php
if (!empty($redirectUrl)) {
    $abs = $urlGenerator->getAbsoluteURL($redirectUrl);
    $redirectUrl = join('?', array_filter(parse_url($abs), fn ($k) => in_array($k, ['path', 'query']), ARRAY_FILTER_USE_KEY));
}
```
which will only take the path+query params of the absolute URL you want to generate.

Wdyt?

julien-nc:
@juliusknorr What do you think about this fix?

bdovaz:
@julien-nc change done!

About the reverse proxy, we have it working on a base path both with nginx and from AWS with ALB without problem for a long time.

The problems have started to occur when switching from `user_saml` to `user_oidc` as I say. As I mentioned, in the link I put `user_saml` already did what I'm doing in this PR so I understand that's why it worked.

In our case the query parameter values arrive without the basePath despite having all the overwrite* set as you say. In fact, obviously if these parameters were set incorrectly we would have more problems than using `user_oidc`.

juliusknorr:
I think the fix makes sense being in line with what user_saml does.

Ideally we'd check on why this is not applied upfront when setting the redirect url, just from reading the code it seems in some cases 
- we pass it through absolute url https://github.com/nextcloud/server/blob/9fcf5a56009e069533ed16131c139bb917b80f72/lib/private/URLGenerator.php#L293
- in others we don't https://github.com/nextcloud/server/blob/9fcf5a56009e069533ed16131c139bb917b80f72/lib/private/URLGenerator.php#L293

juliusknorr (approved):
Haven't tried but looks reasonable (see comment)

julien-nc (approved):

PR #991, opened 2024-11-20
fix: add password confirmation to change admin settings

janepie:


lib/Controller/SettingsController.php
```diff
@@ -73,6 +74,7 @@ public function isDiscoveryEndpointValid($url) {
 		return $result;
 	}
 
+	#[PasswordConfirmationRequired]
```
julien-nc, line 77: `OCP\AppFramework\Http\Attribute\PasswordConfirmationRequired` is only available in NC >= 27. You can use the old annotation instead:

```suggestion
	/**
	 * @PasswordConfirmationRequired
	 */
```

julien-nc (changes requested):

lib/Controller/SettingsController.php
janepie: > `OCP\AppFramework\Http\Attribute\PasswordConfirmationRequired` is only available in NC >= 27. You can use the old annotation instead:

Ah yes, got it, thank you! Not sure about the failing integration checks though, seems to be something with the php version?

lib/Controller/SettingsController.php
```diff
@@ -17,6 +17,7 @@
 use OCP\AppFramework\Controller;
 use OCP\AppFramework\Db\DoesNotExistException;
 use OCP\AppFramework\Http;
+use OCP\AppFramework\Http\Attribute\PasswordConfirmationRequired;
```
julien-nc, line 20: This line should be deleted :wink:.
After having committed this change, you can squash the 3 last commits together with `git rebase -i HEAD~3` and force push.

julien-nc (changes requested):

lib/Controller/SettingsController.php
julien-nc: Yeah we can ignore those until we figure out how to fix them. The wrong version of Php is installed in there.

julien-nc (approved):
Nice interactive rebase!

edward-ly (approved):
Seems fine with me too, although some commit messages adhere to the [Conventional Commits spec](https://www.conventionalcommits.org/en/v1.0.0/) while others don't. Either now or in the future, choose whether you want to adhere to it or not and stick with it.

PR #1002, opened 2024-12-03
Fix: update cache when discovery endpoint is changed

janepie:
Before, endpoint discovery value was cached with ID as key. This keeps the old value in the cache if only the discovery endpoint of a provider is changed in the settings.


julien-nc (approved):
:+1:

PR #1006, opened 2024-12-06
Add support for comma-separated groups in group mapping attribute

julien-nc:
Some IdPs can't provide JSON values in token attributes.


src/components/SettingsForm.vue
```diff
@@ -238,7 +238,7 @@
 			{{ t('user_oidc', 'Use group provisioning.') }}
 		</NcCheckboxRadioSwitch>
 		<p class="settings-hint">
-			{{ t('user_oidc', 'This will create and update the users groups depending on the groups claim in the id token. The Format of the groups claim value should be [{gid: "1", displayName: "group1"}, ...] or ["group1", "group2", ...]') }}
+			{{ t('user_oidc', 'This will create and update the users groups depending on the groups claim in the id token. The Format of the groups claim value should be [{gid: "1", displayName: "group1"}, ...] or ["group1", "group2", ...] or group1,group2') }}
```
janepie, line 241:
```suggestion
			{{ t('user_oidc', 'This will create and update the users groups depending on the groups claim in the id token. The Format of the groups claim value should be [{gid: "1", displayName: "group1"}, ...] or ["group1", "group2", ...] or "group1,group2"') }}
```

janepie (approved):

lib/Service/ProvisioningService.php
```diff
@@ -398,6 +398,11 @@ public function getSyncGroupsOfToken(int $providerId, object $idTokenPayload) {
 		if ($event->hasValue() && $event->getValue() !== null) {
 			// casted to null if empty value
 			$groups = json_decode($event->getValue() ?? '');
+			// support values like group1,group2
+			if (is_string($groups)) {
+				$groups = explode(',', $groups);
+				$groups = array_filter($groups);
```
edward-ly, line 404: In case there is any excess whitespace in the string, maybe it would also be a good idea to use `trim`/`mb_trim` to remove them.
wdyt?

PR #1008, opened 2024-12-10
fix: Handle redirect URL containing a ':'

artonge:
Replace `parse_url` as it does not support path containing a `:`.


julien-nc:
Can it happen with another character than ":"?

artonge:
> Can it happen with another character than ":"?

Probably yes. But we don't have such reports for now.

Thinking about it, I don't think that my solution works, as in case of http://... The initial : will be replaced.

artonge:
Updated to be more generic. This will remove the protocol and the domain name.

julien-nc:
Can you check it does not break the login flow with the desktop client? That was an issue which was fixed by https://github.com/nextcloud/user_oidc/pull/971

PR #1011, opened 2024-12-16
Add global scale (gss) support

julien-nc:
This adds global scale support in user_oidc.
This requires https://github.com/nextcloud/globalsiteselector/pull/168

* Adjust `LoginController::singleLogoutService` to get the provider ID from the GSS logout JWT token
* Adjust the provisioning service to get the user information needed by GSS and store it in the session (uid, email, display name, quota and groups)
* Adjust the user backend: add getUserData (inspired by user_saml) to format the user data for the GSS master

Questions/doubts:

* In `LoginController::singleLogoutService` we try to get the OIDC login token from the session to optionally pass it as a GET parameter to the "end session endpoint" of the OIDC provider. If the session has died on the GSS master, this won't work. We will pass an empty login token. How bad is this?
* We don't support the backchannel logout yet

### DIFF
--- a/lib/Controller/LoginController.php
+++ b/lib/Controller/LoginController.php
@@ -26,6 +26,7 @@ use OCA\UserOIDC\Service\ProviderService;
 use OCA\UserOIDC\Service\ProvisioningService;
 use OCA\UserOIDC\Service\TokenService;
 use OCA\UserOIDC\Vendor\Firebase\JWT\JWT;
+use OCA\UserOIDC\Vendor\Firebase\JWT\Key;
 use OCP\AppFramework\Db\DoesNotExistException;
 use OCP\AppFramework\Db\MultipleObjectsReturnedException;
 use OCP\AppFramework\Http;
@@ -554,6 +555,7 @@ class LoginController extends BaseOidcController {
 	/**
 	 * Endpoint called by NC to logout in the IdP before killing the current session
 	 *
+	 * @PublicPage
 	 * @NoAdminRequired
 	 * @NoCSRFRequired
 	 * @UseSession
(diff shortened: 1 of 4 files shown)


lib/Controller/LoginController.php
```diff
@@ -568,7 +570,22 @@ public function singleLogoutService() {
...
+			if ($isFromGS) {
+				// Request is from master GlobalScale: we get the provider ID from the JWT token provided by the slave
+				$jwt = $this->request->getParam('jwt', '');
+
+				try {
+					$key = $this->config->getSystemValueString('gss.jwt.key', '');
+					$decoded = (array)JWT::decode($jwt, new Key($key, 'HS256'));
+
+					$providerId = $decoded['oidcProviderId'] ?? null;
+				} catch (\Exception) {
```
juliusknorr, line 584: Worth to log at least at info/debug?

lib/User/Backend.php
```diff
@@ -149,6 +149,52 @@ public function getLogoutUrl(): string {
 		);
 	}
 
+	/**
+	 * Return user data from the idp
+	 * Inspired by user_saml
+	 */
+	public function getUserData(): array {
```
juliusknorr, line 156: I don't see this being called, or is it needed from somewhere externally?

juliusknorr: Ah found in https://github.com/nextcloud/globalsiteselector/blob/331f24c43b565ca9a1d4b973a37248b3b563103d/lib/Master.php#L129 for SAML

Could have a cleaner API with something like a IGSUserBackend interface and typed methods to get the needed user data, but that is probably out of scope ;)

juliusknorr (commented):
Some small comments, generally the code looks sane, but I have not tested as discussed.

lib/Controller/LoginController.php
julien-nc: Done

lib/User/Backend.php
julien-nc: That would be nice indeed. We can adjust user_oidc if/when such interface appear in server.

lib/Controller/LoginController.php
```diff
@@ -509,6 +513,9 @@ public function code(string $state = '', string $code = '', string $scope = '',
 			$this->userSession->completeLogin($user, ['loginName' => $user->getUID(), 'password' => '']);
 			$this->userSession->createSessionToken($this->request, $user->getUID(), $user->getUID());
 			$this->userSession->createRememberMeToken($user);
+			// TODO server should/could be refactored so we don't need to manually create the user session and dispatch the login-related events
```
juliusknorr, line 516: Can you file an issue in server and reference that? Helps to keep track of other cases or considerations around that.

juliusknorr (approved):
Code wise this seems fine 👍

PR #1019, opened 2025-01-13
fix: Avoid slow queries in scenarios where we do not need a search

juliusknorr:
Searching users can be slow and we only need it in case of ldap provisioning, so we can skip otherwise. Additionally we can limit the results to make the query possibly faster when it is used.

### DIFF
--- a/lib/Controller/LoginController.php
+++ b/lib/Controller/LoginController.php
@@ -466,18 +466,22 @@ class LoginController extends BaseOidcController {
 		}
 
 		$autoProvisionAllowed = (!isset($oidcSystemConfig['auto_provision']) || $oidcSystemConfig['auto_provision']);
+		$softAutoProvisionAllowed = (!isset($oidcSystemConfig['soft_auto_provision']) || $oidcSystemConfig['soft_auto_provision']);
 
-		// in case user is provisioned by user_ldap, userManager->search() triggers an ldap search which syncs the results
-		// so new users will be directly available even if they were not synced before this login attempt
-		$this->userManager->search($userId);
-		$this->ldapService->syncUser($userId);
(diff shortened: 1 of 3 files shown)


juliusknorr:
Need to adjust to make it stable26 compatible

lib/Controller/LoginController.php
```diff
@@ -467,10 +467,13 @@ public function code(string $state = '', string $code = '', string $scope = '',
 
 		$autoProvisionAllowed = (!isset($oidcSystemConfig['auto_provision']) || $oidcSystemConfig['auto_provision']);
 
-		// in case user is provisioned by user_ldap, userManager->search() triggers an ldap search which syncs the results
-		// so new users will be directly available even if they were not synced before this login attempt
-		$this->userManager->search($userId);
-		$this->ldapService->syncUser($userId);
+		if (!$this->provisioningService->hasOidcUserProvisitioned($userId) && $this->ldapService->isLDAPEnabled()) {
```
julien-nc, line 470: Not sure this condition is enough. What if the user exists in the user_oidc backend but auto-provisioning is disabled and we still expect to find the user via LDAP. We need to force-sync then too. Maybe this exact case never happens but maybe we are missing some other cases.

IMO We have to make sure to trigger those when (auto-provisioning is disabled) OR (soft-autoprovisioning is enabled AND the user is not found in the user_oidc backend).
In other words: In all the cases where we might need to find the user in another backend.

julien-nc (commented):
Looks good.

lib/Controller/LoginController.php
juliusknorr: All right, then lets add both cases, is still fine with me for the case I'm looking at where ldap is not used

julien-nc (approved):

juliusknorr:
Restarted the test as it seems like something odd on the runner

PR #1039, opened 2025-02-05
Remove AZP check when validating a bearer ID token

### LINKED ISSUE #1038: No azp validation should happen for Bearer Token authentication
### How to use GitHub

---

### Steps to reproduce

The scenario is that we have two applications (Nextcloud and another app), that share the same Identity provider (IDP). The IDP is OIDC compliant and follows relevant RFCs. The other app (let's call it "API-Client") wants to make API calls to Nextcloud, using Access Tokens it obtained from the IDP.

Steps:

1. Configure Nextcloud and API-Client to use a Keycloak as their IDP
2. Configure Nextcloud to accept Access Tokens issued by that IDP for API requests (this is possible via UI)
3. Make sure the API-Client obtains a user-bound access token with Nextcloud in the audience
    * e.g. this can be done through Token Exchange or by isssuing Access Tokens that contain a wider audience from the beginning
4. Perform an API request to Nextcloud

### Expected behaviour
Since we obtained an access token and the audience of the access token includes our Nextcloud, the token should be accepted.

### Actual behaviour
By default Nextcloud will validate that the `azp` claim is equal to Nextcloud's own Client ID, which will make validation of the access token fail.

As far as I can tell, there is no specification covering this behaviour and IMO the validation should be removed entirely, but I'd also find it acceptable to change the default for `selfencoded_bearer_validation_azp_check` to `false`.

### Explanation

This is kind of a follow-up to #856 and #918. Let me try to explain why I think the current behaviour is a bug, by quoting the relevant specifications.

As far as I can tell, the `azp` check was introduced because of [OIDC Core 1.0, Section 3.1.3.7](https://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation):

> Clients MUST validate the ID Token in the Token Response in the following manner: 
> [...]
> 5. This validation MAY include that when an `azp` (authorized party) Claim is present, the Client SHOULD verify that its `client_id` is the Claim Value.

I want to point out that this piece of the specification aims at validating an **ID Token**, not an access token. Therefore I think the `azp` check as implemented in the `LoginController` (setting `login_validation_azp_check`) **is correct** and should continue to default to true. Because in that case NC obtains an ID token from the IDP itself and of course it can expect that the IDP made that token for NC and no other party.

However, this is not the case that we have when validating the Bearer Token during an API request. First and foremost, **ID Tokens are not access tokens**. [OIDC Core 1.0 specifies that](https://openid.net/specs/openid-connect-core-1_0.html#Terminology) it is using the term "Access Token" as defined in [RFC 6749 (OAuth 2.0)](https://www.rfc-editor.org/rfc/rfc6749.html). This means OIDC itself has no expectations about the access token at all, not even that it might be a JWT, because the original OAuth 2.0 spec [explicitly makes no such assumptions](https://www.rfc-editor.org/rfc/rfc6749.html#section-1.4):

> The token may denote an identifier used to retrieve the authorization information or may self-contain the authorization information in a verifiable manner (i.e., a token string consisting of some data and a signature).  Additional authentication credentials, which are beyond the scope of this specification, may be required in order for the client to use a token.

As such, you will not find the term `azp` anywhere in the OAuth 2.0 specs and (as said before) the OIDC spec doesn't change anything about access tokens. There is also a relatively new spec (2021) called [RFC 9068](https://www.rfc-editor.org/rfc/rfc9068) that defines a JWT profile for OAuth 2.0 Access Tokens. It essentially collects best practices established so far and clarifies them. This spec also reminds of the importance to not mix up ID Tokens and Access Tokens:

> See the [Security Considerations](https://www.rfc-editor.org/rfc/rfc9068#SecurityConsiderations) section for details on the importance of preventing OpenID Connect ID Tokens (as defined by Section 2 of [[OpenID.Core](https://www.rfc-editor.org/rfc/rfc9068#OpenID.Core)]) from being accepted as access tokens by resource servers implementing this profile.

Apart from that, it is probably the first spec to clarify [how to validate a JWT Access Token](https://www.rfc-editor.org/rfc/rfc9068#name-validating-jwt-access-token)\* (which was previously left up to "commercial implementations" to agree on). It confirms that validating the audience [is the right thing to do](https://www.rfc-editor.org/rfc/rfc9068#section-4-5.4):

> The resource server MUST validate that the "aud" claim contains a resource indicator value corresponding to an identifier the resource server expects for itself. The JWT access token MUST be rejected if "aud" does not contain a resource indicator of the current resource server as a valid audience.

However, also this spec doesn't cover `azp`. In general, when searching the web I can't find any specifications apart from OIDC 1.0 Core that mention the `azp` claim at all. The original [JWT spec](https://www.rfc-editor.org/rfc/rfc7519#section-4.1.3) mentions the `aud` claim, but has nothing on `azp`. On the other hand, seeing what `azp` is used for, it becomes clear that it identifies the party that requested the token. Thus it makes sense to validate that `azp == client_id` if you are sure that you requested the token yourself, but this also means it makes no sense to perform the same validation in cases where you can be sure that your application was **not** the application that requested the token, which is the case in Bearer Token authentication.

I hope this explanation was somewhat helpful. To summarize:
* There is no spec telling us to validate `azp` on access tokens
* The only spec mentioning `azp` does so in a context of self-obtained ID Tokens
    * that spec also says `azp` is "the party to which the ID Token was issued", i.e. it wouldn't make sense to be the current client ID, unless Nextcloud is making requests to itself
* Two specs agree that the `aud` claim is what you want to check to see whether a token is intended for your application

\* Important note: While RFC 9068 explains how to validate an access token, this includes a _new_ check on the `typ` header. I'd currently advise against implementing such a check, because implementations such as KeyCloak did not yet catch up on this specification and indicate an ambiguous type in their header's still. IMO this needs to be adopted by IDPs first, before resource servers should validate it.

julien-nc:
closes #1038

As explained in #1038, checking the AZP (authorized party) only makes sense when validating an ID token that is for ourselves (during login, in our context). So let's remove it from `SelfEncodedValidator::isValidBearerToken`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -291,14 +291,14 @@ sudo -u www-data php /var/www/nextcloud/occ user_oidc:provider demoprovider \
 
 ### Disable audience and azp checks
 
-The `audience` and `azp` token claims will be checked when validating a login or bearer ID token.
-You can disable these check with these config value (in config.php):
+The `audience` and `azp` token claims will be checked when validating a login ID token.
+Only the `audience` will be checked when validating a Bearer token.
+You can disable these checks with these config values (in config.php):
 ``` php
 'user_oidc' => [
     'login_validation_audience_check' => false,
     'login_validation_azp_check' => false,
     'selfencoded_bearer_validation_audience_check' => false,
-    'selfencoded_bearer_validation_azp_check' => false,
 ],
 ```
 
--- a/lib/User/Validator/SelfEncodedValidator.php
+++ b/lib/User/Validator/SelfEncodedValidator.php
@@ -57,6 +57,7 @@ class SelfEncodedValidator implements IBearerTokenValidator {
 		}
 
 		$oidcSystemConfig = $this->config->getSystemValue('user_oidc', []);
+		// ref https://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation
 		$checkAudience = !isset($oidcSystemConfig['selfencoded_bearer_validation_audience_check'])
 			|| !in_array($oidcSystemConfig['selfencoded_bearer_validation_audience_check'], [false, 'false', 0, '0'], true);
 		$providerClientId = $provider->getClientId();
@@ -67,17 +68,6 @@ class SelfEncodedValidator implements IBearerTokenValidator {
 					|| (is_array($tokenAudience) && !in_array($providerClientId, $tokenAudience, true))
 			) {
 				$this->logger->debug('This token is not for us, the audience does not match the client ID');
-				return null;
-			}
-		}
-
-		$checkAzp = !isset($oidcSystemConfig['selfencoded_bearer_validation_azp_check'])
-			|| !in_array($oidcSystemConfig['selfencoded_bearer_validation_azp_check'], [false, 'false', 0, '0'], true);
-		if ($checkAzp) {
-			// ref https://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation
-			// If the azp claim is present, it should be the client ID
-			if (isset($payload->azp) && $payload->azp !== $providerClientId) {
-				$this->logger->debug('This token is not for us, authorized party (azp) is different than the client ID');
 				return null;
 			}
 		}


julien-nc:
Merged as it's not a risky change.

PR #1040, opened 2025-02-05
Add support for bearer token validation by the OIDC Identity Provider app via en event

julien-nc:
Add optional extra bearer token validation using the [OIDC Identity Provider app](https://github.com/H2CK/oidc) 
See event definition: https://github.com/H2CK/oidc/pull/515
This PR depends on https://github.com/H2CK/oidc/pull/521

If this validation is enabled in user_oidc, it happens before the classic selfencoded and userinfo validations.

This allows clients and external services to use an Oidc access token they would have obtained by using Nextcloud as an IdP (via the OIDC Identity Provider app) as a bearer token to make API requests to Nextcloud.


saw-jan:
I tried this branch with integration_openproject app, I can get the access token but while using token to request OpenProject api (`https://openproject.local/api/v3/users/me`), I get `401 Unauthorized` response. Am I missing something?
```bash
curl 'https://openproject.local/api/v3/users/me' \
-H"Authorization: Bearer wUXTA3RoumzI9BpV5RCFXj6WixfDlmKI8DchQwuK24VKig6n3PxAUBNfRz6XIYupgQk3JOU0" -v
```

```
oidc version: 1.4.0
```

`config.php`
```php
'user_oidc' => 
  array (
    'oidc_provider_token_generation' => true,
    'oidc_provider_bearer_validation' => true,
  ),
```

NobodysNightmare:
This is apparently caused by the access token not being a JWT, but opaque. @julien-nc is the corresponding change on the NC side already merged? Do we have to configure anything to make use of it?

saw-jan:
> This is apparently caused by the access token not being a JWT, but opaque. @julien-nc is the corresponding change on the NC side already merged? Do we have to configure anything to make use of it?

So that means oidc app should generate JWT tokens which then can be used to request OpenProject APIs

julien-nc:
@saw-jan You get 2 tokens in the event data:
* access_token which is not a JWT one and can only be used to reach NC's API (like you did to reach `/apps/oidc/userinfo`)
* id_token which is a JWT

You can use the id_token.

@NobodysNightmare The fact that the access_token is opaque won't change. In the `oidc` app's implementation, it is a randomly generated string. According to the OAuth2 specs: `The string is usually opaque to the client` and `Access tokens can have different formats, structures, and methods of utilization`.

https://datatracker.ietf.org/doc/html/rfc6749#section-1.4

saw-jan:
>@saw-jan You get 2 tokens in the event data:
>
>- access_token which is not a JWT one and can only be used to reach NC's API (like you did to reach /apps/oidc/userinfo)
>- id_token which is a JWT

Ahh, Thanks @julien-nc. I will give it a try again with id_token

saw-jan:
Tried with id-token as well.
```
curl 'https://openproject.local/api/v3/users/me' -H"Authorization: Bearer eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1NiIsImtpZCI6IjAyMzNmYWNlLWQ5NjYtNDI0My05N2QxLTZjMGVmNWQ5ZmEzMCJ9.eyJpc3MiOiJodHRwczpcL1wvbmV4dGNsb3VkLmxvY2FsIiwic3ViIjoiYWRtaW4iLCJhdWQiOiJNRndJMVdPaGswWEJERGhkNmg4aU1JZ0RuQXh1TFpCeHNGTEhDVHdxa0NqU0w4U0kyUG5EQmhYdGJKVk0zM2lEIiwiZXhwIjoxNzQxMDAzNjcwLCJhdXRoX3RpbWUiOjE3NDEwMDI3NzAsImlhdCI6MTc0MTAwMjc3MCwiYWNyIjoiMCIsImF6cCI6Ik1Gd0kxV09oazBYQkREaGQ2aDhpTUlnRG5BeHVMWkJ4c0ZMSENUd3FrQ2pTTDhTSTJQbkRCaFh0YkpWTTMzaUQiLCJwcmVmZXJyZWRfdXNlcm5hbWUiOiJhZG1pbiIsInNjb3BlIjoib3BlbmlkIHByb2ZpbGUgZW1haWwgcm9sZXMiLCJuYmYiOjE3NDEwMDI3NzAsImp0aSI6IjYiLCJyb2xlcyI6WyJhZG1pbiJdLCJ1cGRhdGVkX2F0IjoxNzQxMDAyNDI0LCJuYW1lIjoiYWRtaW4iLCJmYW1pbHlfbmFtZSI6ImFkbWluIiwiZ2l2ZW5fbmFtZSI6IiIsIm1pZGRsZV9uYW1lIjoiIn0.i9usnGr7lyCgwHI3_EfYnM6yxK_Vn83X2WX93XLJxilu-g0xgTsFmtoNbzbV-m-ffTrP1vGEav3SoWxpTizup1me3ltjAREzeKekoK3wZ1d4k0LM5hX2iw2_UjnpwRy6qVB2nzu57JuM5nsuaIDZEtVJLsc2mPtwSR7qGpLCV-fCKC445M3JS-TMxM1E_qbqytpCOws24D8TcgOA6yTTaoYLlXsYWLS5RSYEek4nxiR4yTeHXcaQXR9g45EvBDyoOHJdmtEqTQxd9NfQh7fgDKFMj37_7zn2BWudDdrfjPjQHSka8bKPDt_RBDxhVv30ZaUv6SeCVuA_vcQeGnSzSMa8biaND4qkSRMbF3LYWsmAmQeDA-gxfN04bmtN8kMsVZOjV67OR4H_uR5Zko30ysPVUm2Egwd5IWfeC2GevmjxMRrLr40uQ4uITjRj0HrZ7SoqCBz1LsR4NjZBm1U72vz23AwdyjbRoWDCWrYrH8WAfjM-hbKVvvQps7afld4caWHgKPqKclSCdnmFs-6dROgWXo2tiraijo57c1Wm-lTLDvGfjbFBHvoSqnNLN7r6NOVcK3ZPby9wmmykxescRKviZ52DTVIgkSK2O7xGWxxhRJ5iU-knSbJFKW-7bsMdTrkOIiG8beONn1P-UzOWML_qSlBvmOuvaOYzwHwDt4Y" -v
```
but no luck. OP still returns `401`

saw-jan:
> Tried with id-token as well.
> 
> ```
> curl 'https://openproject.local/api/v3/users/me' -H"Authorization: Bearer eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1NiIsImtpZCI6IjAyMzNmYWNlLWQ5NjYtNDI0My05N2QxLTZjMGVmNWQ5ZmEzMCJ9.eyJpc3MiOiJodHRwczpcL1wvbmV4dGNsb3VkLmxvY2FsIiwic3ViIjoiYWRtaW4iLCJhdWQiOiJNRndJMVdPaGswWEJERGhkNmg4aU1JZ0RuQXh1TFpCeHNGTEhDVHdxa0NqU0w4U0kyUG5EQmhYdGJKVk0zM2lEIiwiZXhwIjoxNzQxMDAzNjcwLCJhdXRoX3RpbWUiOjE3NDEwMDI3NzAsImlhdCI6MTc0MTAwMjc3MCwiYWNyIjoiMCIsImF6cCI6Ik1Gd0kxV09oazBYQkREaGQ2aDhpTUlnRG5BeHVMWkJ4c0ZMSENUd3FrQ2pTTDhTSTJQbkRCaFh0YkpWTTMzaUQiLCJwcmVmZXJyZWRfdXNlcm5hbWUiOiJhZG1pbiIsInNjb3BlIjoib3BlbmlkIHByb2ZpbGUgZW1haWwgcm9sZXMiLCJuYmYiOjE3NDEwMDI3NzAsImp0aSI6IjYiLCJyb2xlcyI6WyJhZG1pbiJdLCJ1cGRhdGVkX2F0IjoxNzQxMDAyNDI0LCJuYW1lIjoiYWRtaW4iLCJmYW1pbHlfbmFtZSI6ImFkbWluIiwiZ2l2ZW5fbmFtZSI6IiIsIm1pZGRsZV9uYW1lIjoiIn0.i9usnGr7lyCgwHI3_EfYnM6yxK_Vn83X2WX93XLJxilu-g0xgTsFmtoNbzbV-m-ffTrP1vGEav3SoWxpTizup1me3ltjAREzeKekoK3wZ1d4k0LM5hX2iw2_UjnpwRy6qVB2nzu57JuM5nsuaIDZEtVJLsc2mPtwSR7qGpLCV-fCKC445M3JS-TMxM1E_qbqytpCOws24D8TcgOA6yTTaoYLlXsYWLS5RSYEek4nxiR4yTeHXcaQXR9g45EvBDyoOHJdmtEqTQxd9NfQh7fgDKFMj37_7zn2BWudDdrfjPjQHSka8bKPDt_RBDxhVv30ZaUv6SeCVuA_vcQeGnSzSMa8biaND4qkSRMbF3LYWsmAmQeDA-gxfN04bmtN8kMsVZOjV67OR4H_uR5Zko30ysPVUm2Egwd5IWfeC2GevmjxMRrLr40uQ4uITjRj0HrZ7SoqCBz1LsR4NjZBm1U72vz23AwdyjbRoWDCWrYrH8WAfjM-hbKVvvQps7afld4caWHgKPqKclSCdnmFs-6dROgWXo2tiraijo57c1Wm-lTLDvGfjbFBHvoSqnNLN7r6NOVcK3ZPby9wmmykxescRKviZ52DTVIgkSK2O7xGWxxhRJ5iU-knSbJFKW-7bsMdTrkOIiG8beONn1P-UzOWML_qSlBvmOuvaOYzwHwDt4Y" -v
> ```
> 
> but no luck. OP still returns `401`
> 
> https://github.com/nextcloud/user_oidc/blob/7ddd8160ce94fd6bae8e9b69a25b24d2ee25cc66/lib/Model/Token.php#L37-L38

my BAD, I forgot to enable `oidc_provider_bearer_validation` config. Now it is working :+1:

julien-nc:
@saw-jan Hmmm isn't OP supposed to validate the JWT token without making any request to NC?
I'm asking because you mention that enabling `oidc_provider_bearer_validation` on the NC side helped to make a request to OP.

@NobodysNightmare I agree with you.
Short term, the oidc app will keep producing opaque access tokens. Of course we can discuss that but since using ID tokens can satisfy our use case, let's first go with that. My influence on the oidc app is limited :grin: .

saw-jan:
>@saw-jan Hmmm isn't OP supposed to validate the JWT token without making any request to NC?
I'm asking because you mention that enabling oidc_provider_bearer_validation on the NC side helped to make a request to OP.

I have no idea how OP validates the received token but AFAIK resource server (OP) needs to validate with auth server (NC).
But my knowledge is limited here. May be @NobodysNightmare could shed some light here :smiley:

NobodysNightmare:
Sorry, I forgot to respond here. To be honest, I also don't know why changing this option would make a difference.

We can experiment locally if you want. But to my knowledge, OpenProject makes no authenticated requests to Nextcloud to validate a JWT. The only request that comes to my mind is fetching the signing keys (JWKS), but that endpoint should not require authentication of any kind.

lib/Service/TokenService.php
```diff
@@ -311,4 +315,44 @@ public function getExchangedToken(string $targetAudience): Token {
...
+
+	/**
+	 * Try to get a token from the Oidc provider app for a user and a specific audience (client ID)
+	 *
+	 * @param string $userId
+	 * @param string $targetAudience
+	 * @return Token|null
+	 */
+	public function getTokenFromOidcProviderApp(string $userId, string $targetAudience): ?Token {
+		if (!class_exists('OCA\OIDCIdentityProvider\AppInfo\Application')) {
```
juliusknorr, line 327:
```suggestion
		if (!class_exists(\OCA\OIDCIdentityProvider\AppInfo\Application::class)) {
```

This is fine in PHP even if the app is not loaded / available.

lib/User/Backend.php
```diff
@@ -229,6 +231,24 @@ public function getCurrentUserId(): string {
 			}
 		}
 
+		// check if we should ask the OIDC Identity Provider app (app_id: oidc) to validate the token (default is false)
+		if ($ncOidcProviderBearerValidation) {
+			if (class_exists('OCA\OIDCIdentityProvider\Event\TokenValidationRequestEvent')) {
+				$validationEvent = new \OCA\OIDCIdentityProvider\Event\TokenValidationRequestEvent($headerToken);
+				$this->eventDispatcher->dispatchTyped($validationEvent);
```
juliusknorr, line 238: Since this calls external code we cannot be sure about that it remains unchanged. I'm not a fan API wise but cannot think of a better way.

I would wrap this with try/catch generally on any Throwable to avoid bringing down an instance where any issue occurs here

lib/Service/TokenService.php
```diff
@@ -311,4 +315,44 @@ public function getExchangedToken(string $targetAudience): Token {
...
+			$this->logger->warning('[TokenService] Failed to get token from Oidc provider app, oidc app is not enabled');
+			return null;
+		}
+		if (!class_exists('OCA\OIDCIdentityProvider\Event\TokenGenerationRequestEvent')) {
+			$this->logger->warning('[TokenService] Failed to get token from Oidc provider app, TokenGenerationRequestEvent class not found, oidc app is probably not >= v1.4.0');
+			return null;
+		}
+
+		$generationEvent = new \OCA\OIDCIdentityProvider\Event\TokenGenerationRequestEvent($targetAudience, $userId);
+		$this->eventDispatcher->dispatchTyped($generationEvent);
```
juliusknorr, line 341: Same here for a try/catch I'd say

juliusknorr (approved):
Small comments, otherwise seems ok

julien-nc:
@saw-jan @NobodysNightmare This was released with `user_oidc` v7.0.0!
The `oidc` app >= v1.4.0 is required. With v1.4.0, the refresh tokens expires when the token expires. With the next release (probably v1.5.0), it will be possible to configure/disable the refresh token expiration.

PR #1041, opened 2025-02-05
Call userinfo on login to enrich the login ID token

### LINKED ISSUE #1034: Provision groups from userinfo/introspection endpoints information
### How to use GitHub

---

## Feature request

30.0.5

Some users in our keycloak instance have a lot of groups/roles and because of that we had issues with too large id tokens.
Therefore we only provide that information via userinfo/introspection endpoints.
We are in the process of migrating multiple of services to keycloak authentication and Nextcloud is one of them.
So far this app is very promising, but we cannot sync our groups because this app does not use userinfo/introspection for that.

It would be great if user_oidc could fetch group information from userinfo/introspection endpoints.

One could increase the maximum header size setting in i.e. nginx/nextcloud and try to use id token, but that is not a bullet-proof solution because the id tokens might grow even more in the future.

**Additional context**
No additional context.

julien-nc:
closes #1034

If, for some reason, some user information is missing from the login ID tokens but can be obtained with the userinfo endpoint, user_oidc can now optionally call the userinfo endpoint on login to enrich the data contained in the login ID token.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,21 @@ To change this behaviour and disable the default claims, you can change this val
 When default claims are disabled, each claim will be asked for only if there is an attribute explicitely mapped
 in the OpenId client settings (in Nextcloud's admin settings).
 
+### Call the userinfo endpoint to enrich the login ID token
+
+If some user information is not in your login ID tokens but can be obtained with the userinfo endpoint, just enable
+`enrich_login_id_token_with_userinfo` in config.php. This is disabled by default.
+``` php
+'user_oidc' => [
+    'enrich_login_id_token_with_userinfo' => true,
+],
+```
+
+This will use the content of the userinfo endpoint response just like if it had been included in the login ID token.
+
+This will only work on login and not when validating a bearer token
+because provisioning when validating a bearer access token is not supported yet.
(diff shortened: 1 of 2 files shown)


jvllmr:
I mentioned this on my issue before, but this change works well for me in my test setup ✅

hendrik1120:
> I'm not a fan of all the extra settings, so mostly asking if we know a reason why this is not sent right away.

@juliusknorr Yes, there is a pretty good reason. The claims parameter allows the user to granularly select which information should be shared with the relying party during consent.

If user_oidc is configured to only request the `openid` scope and the following claims parameter:
```json
{
  "userinfo": {
    "preferred_username": {"essential":  true},
    "role": null,
    "organisation": null
  }
}
```
The user can opt out of the non-essential claims individually. This makes a lot of sense for privacy.
This is how it looks using Authelia 4.39:

Bildschirmfoto 2025-03-17 um 22 51 43

However, user_oidc currently does not support marking individual claims as required, so this could be abused to exclude the set quota in which nextcloud falls back to the default quota.

juliusknorr (approved):

PR #1042, opened 2025-02-05
Optionally disable ssl checks in all client requests

### LINKED ISSUE #988: Disable ssl verification in development
### How to use GitHub

---

## Feature request

30.0.2

When we are developing there is no way to run Keycloak and Nextcloud over https with a self-signed certificate and have the user_oidc app work correctly. There are several scripts that make requests that do not take into account the ability to set the `verify` field of Guzzle:

> Search for `$client->` inside this scripts

https://github.com/nextcloud/user_oidc/blob/main/lib/Controller/LoginController.php
https://github.com/nextcloud/user_oidc/blob/main/lib/Service/DiscoveryService.php

Some of the work is already done because at least for certain errors it is taken into account that if Nextcloud is in debug mode it considers that it is safe:

https://github.com/nextcloud/user_oidc/blob/d161cdd8b254ca90246b362165003ae2854cec1e/lib/Controller/LoginController.php#L87

Nextcloud does not have the option to globally disable verify by default, which would be ideal:

https://github.com/nextcloud/server/blob/d334773b9868561db86209bbf801d2835f182a7b/lib/private/Http/Client/Client.php#L55

Assume that if Nextcloud is in debug mode it sends on Guzzle requests: $options['verify'] = false

Or have a setting in config.php itself as some apps do:

https://github.com/nextcloud/richdocuments/blob/05fa51e71a408a16624588d4f7aed23257090bb2/lib/Service/CachedRequestService.php#L95

**Additional context**

julien-nc:
closes #988

Can be useful when using self signed certs for the IdP.

### DIFF
--- a/README.md
+++ b/README.md
@@ -296,6 +296,16 @@ This app can stop matching users (when a user search is performed in Nextcloud) 
 ],
 ```
 
+### Disable SSL certificates verification
+
+If your IdP is using self-signed certificates or if you have any other reason to disable SSL certificates verification
+when making requests to the IdP, set this in config.php :
+``` php
+'user_oidc' => [
+    'disable_certificate_verification' => true,
+],
+```
+
 ## Building the app
 
 Requirements for building:
--- a/lib/Controller/Id4meController.php
+++ b/lib/Controller/Id4meController.php
@@ -17,6 +17,7 @@ use OCA\UserOIDC\Db\Id4MeMapper;
 use OCA\UserOIDC\Db\UserMapper;
 use OCA\UserOIDC\Helper\HttpClientHelper;
 use OCA\UserOIDC\Service\ID4MeService;
+use OCA\UserOIDC\Service\NetworkService;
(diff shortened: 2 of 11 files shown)


juliusknorr:
Isn't it feasible to just import the self signed ca into the certificate chain? That way you could still validate the certificate

saw-jan:
Similar to what we do in [integration_openproject](https://github.com/nextcloud/integration_openproject). import the certs in development environment to be able to use self-signed certs.
See https://github.com/nextcloud/integration_openproject/tree/master/dev

juliusknorr:
I'm fine either way, we have this kind of setting in a lot of places in Nextcloud.

bdovaz:
Any updates? It's been inactive for months.

Thank you.

bdovaz:
Any updates? It's been inactive for months.

Thank you.

cc @julien-nc @juliusknorr

PR #1046, opened 2025-02-06
Map the user language

### LINKED ISSUE #987: Map the Nextcloud language field
### How to use GitHub

---

## Feature request

30.0.2

Right now there is no way to map the Nextcloud language with an attribute.

Possibility of mapping in settings as another attribute and modifying the Nextcloud user language and propagating it to Nextcloud so that the language is always synchronized from a single point.

**Additional context**

julien-nc:
* Add attribute mapping for the user account language
* Adjust Upsert command
* Adjust admin settings UI

closes #987


lib/Service/ProvisioningService.php
```diff
@@ -230,6 +235,19 @@ public function provisionUser(string $tokenUserId, int $providerId, object $idTo
 			$account->setProperty('phone', $event->getValue(), $scope, '1', '');
 		}
 
+		$event = new AttributeMappedEvent(ProviderService::SETTING_MAPPING_LANGUAGE, $idTokenPayload, $language);
+		$this->eventDispatcher->dispatchTyped($event);
+		$this->logger->debug('Phone mapping event dispatched');
+		if ($event->hasValue()) {
+			$language = $event->getValue();
+			$languagesCodes = $this->l10nFactory->findAvailableLanguages();
+			if (in_array($language, $languagesCodes, true) || $language === 'en') {
```
juliusknorr, line 244: Why is there a special case for en?

juliusknorr (approved):
Small question but otherwise looks sane 👍

lib/Service/ProvisioningService.php
```diff
@@ -230,6 +235,19 @@ public function provisionUser(string $tokenUserId, int $providerId, object $idTo
 			$account->setProperty('phone', $event->getValue(), $scope, '1', '');
 		}
 
+		$event = new AttributeMappedEvent(ProviderService::SETTING_MAPPING_LANGUAGE, $idTokenPayload, $language);
+		$this->eventDispatcher->dispatchTyped($event);
+		$this->logger->debug('Phone mapping event dispatched');
```
saw-jan, line 240: should be language?
```suggestion
		$this->logger->debug('Language mapping event dispatched');
```

lib/Service/ProvisioningService.php
julien-nc: Because I checked how the language is set by the settings app with `PUT /ocs/v2.php/cloud/users/USER_ID` : https://github.com/nextcloud/server/blob/master/apps/provisioning_api/lib/Controller/UsersController.php#L1080-L1084
to know which values are valid and did the same.

lib/Service/ProvisioningService.php
julien-nc: Thanks

PR #1049, opened 2025-02-09
fix: backchannel logout token may not contain "sub"

prigaux:
The spec says:

> A Logout Token MUST contain either a sub or a sid Claim, and MAY contain both

Ory Hydra OP never sends "sub".

The fix is really trivial: ignore "sub" if it is missing.


julien-nc (approved):
The specs are actually more complex than what i described. It's a good first step to support the absence of the `sub`. Thanks!

PR #1058, opened 2025-02-17
fix(ProvisioningService): Handle `InvalidArgumentException` when updating account

susnux:
The `updateAccount` is documented as throwing an `InvalidArgumentException` if one of the changed properties is invalid. We should then simply reset that value.


nickvergessen (approved):

julien-nc (approved):
Tested with an invalid phone number, works fine!

PR #1061, opened 2025-02-18
Fix password confirmation

julien-nc:
refs https://github.com/nextcloud/server/issues/43612

1. Replicate what's done in user_saml's user backend with the session: Set `last-password-confirm` in the session everytime the user ID is obtained in `OCA\UserOIDC\User\Backend::getCurrentUserId()`
2. Adjust auth token scope in login controller. Inspired by https://github.com/nextcloud/server/pull/43942/files#diff-b30a8c4cef5da5cede4d9038c16ede43e1746f85270e4249598fd305b0fa77deR173-R176
3. Bump min NC version to 28 to make sure we have `OCP\Authentication\Token\IToken`
4. Only do 2. if `IToken::SCOPE_SKIP_PASSWORD_VALIDATION` is defined (it was introduced in 30)

Maybe 1. is not necessary. Wdyt?

### DIFF
--- a/lib/Controller/LoginController.php
+++ b/lib/Controller/LoginController.php
@@ -36,6 +36,7 @@ use OCP\AppFramework\Http\JSONResponse;
 use OCP\AppFramework\Http\RedirectResponse;
 use OCP\AppFramework\Http\TemplateResponse;
 use OCP\AppFramework\Utility\ITimeFactory;
+use OCP\Authentication\Token\IToken;
 use OCP\DB\Exception;
 use OCP\EventDispatcher\IEventDispatcher;
 use OCP\Http\Client\IClientService;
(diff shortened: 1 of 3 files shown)


waza-ari:
Just for confirmation, `SCOPE_SKIP_PASSWORD_VALIDATION` is just a constant, nothing that needs to be changed from a user perspective, correct?

julien-nc:
@waza-ari It is internal to the server. Nothing that the users or the admins should be aware of.

lib/Controller/LoginController.php
```diff
@@ -520,6 +521,14 @@ public function code(string $state = '', string $code = '', string $scope = '',
 			$this->userSession->completeLogin($user, ['loginName' => $user->getUID(), 'password' => '']);
 			$this->userSession->createSessionToken($this->request, $user->getUID(), $user->getUID());
 			$this->userSession->createRememberMeToken($user);
+
+			// prevent password confirmation
+			if (defined(IToken::class . '::SCOPE_SKIP_PASSWORD_VALIDATION')) {
+				$token = $this->authTokenProvider->getToken($this->session->getId());
+				$token->setScope([IToken::SCOPE_SKIP_PASSWORD_VALIDATION => true]);
```
blizzz, line 528: IIRC, if another scope was set, it needs to be taken over, this would replace it.

blizzz (commented):
One thing to double check

lib/Controller/LoginController.php
julien-nc: Thanks for the feedback.
Hmmm are you saying we should override the scope and just set `IToken::SCOPE_SKIP_PASSWORD_VALIDATION` to true?
Or are you saying we should only extend the scope and add `IToken::SCOPE_SKIP_PASSWORD_VALIDATION`?

juliusknorr (approved):
Looks good with the additional commit addressing the earlier review comment 👍

lib/Controller/LoginController.php
blizzz: >  Or are you saying we should only extend the scope and add IToken::SCOPE_SKIP_PASSWORD_VALIDATION?

This one!

PR #1065, opened 2025-02-24
broadcast prelogin event before login event

ArtificialOwl:
The fix the fact that, on global scale, a session is authed on master.


julien-nc (approved):
Makes sense!

Issue #1091, opened 2025-03-31
Support requesting extra scopes when using token-requested events


saw-jan:
## Feature request
Support requesting extra scopes when using token-requested events (`InternalTokenRequestedEvent`, `ExternalTokenRequestedEvent`, `ExchangedTokenRequestedEvent`).

**Describe the solution you'd like**
I would like to request an option to pass the scopes when instantiating token-requested event. When the event is received, the authorization request would then include the requested scopes.

```php
$scopes = ["api_v3"];
$event = new ExchangedTokenRequestedEvent($scopes);
$this->eventDispatcher->dispatchTyped($event);
```

**Additional context**
For `InternalTokenRequestedEvent`, the implementation needs to be done first in the `oidc` app (See https://github.com/H2CK/oidc/issues/537)
And regarding `ExternalTokenRequestedEvent` if the requested scope is optional, the first login token might not contain that optional scope. Therefore, when this event is called with scopes, the token needs to be refreshed or something needs to be done to update the token store with the token that includes the requested scope.

NOTE:
For `ExternalTokenRequestedEvent` and `ExchangedTokenRequestedEvent`, we can add the required scopes while adding the OIDC provider but that would mean every token will contain those scopes. If we have the option to specify scopes while calling the events, then we would be able to control the optional scopes to request when needed, preventing unnecessary scopes in the tokens.

## [UPDATED USAGE]
```php
$scopes = "myscope1 myscope2";
$resource = $targetAudience;

$event1 = new InternalTokenRequestedEvent($targetAudience, $scopes, $resource);

$event2 = new ExchangedTokenRequestedEvent($targetAudience, $scopes);

$event2 = new ExternalTokenRequestedEvent($scopes);
```

saw-jan:
CC @NobodysNightmare

saw-jan:
Implementation in oidc app has been done in PR https://github.com/H2CK/oidc/pull/538
What we need to move forward:
1. In `oidc` app:
    - ability to request extra scopes using `TokenGenerationRequestEvent`
    - `TokenGenerationRequestListener` needs to pass the requested scopes properly
2. In `user_oidc` app, `ExchangedTokenRequestedEvent`, `ExternalTokenRequestedEvent` and `InternalTokenRequestedEvent` needs to be able to accept extra scopes params.

saw-jan:
Necessary implementations in odic app:
- https://github.com/H2CK/oidc/pull/541
[...]

PR #1098, opened 2025-04-10
feat: allow requesting scopes when using InternalTokenRequestedEvent event

saw-jan:
Allow requesting `scope` and `resource` when using `InternalTokenRequestedEvent` event.
Necessary implemention in oidc app:
- https://github.com/H2CK/oidc/pull/541
- https://github.com/H2CK/oidc/pull/542

Usage:
```php
$scopes = ["myscope1", "myscope2"];
$resource = $targetAudience;

$event = new InternalTokenRequestedEvent($targetAudience, $scopes, $resource);
```


julien-nc (commented):
Makes sense.
Same as #1099 , extra scopes could be an array.

saw-jan:
> Makes sense. Same as #1099 , extra scopes could be an array.

Done!

julien-nc (approved):

saw-jan:
@julien-nc Please merge this if okay to do so. I have no permissions to merge it

PR #1099, opened 2025-04-10
feat: allow requesting scopes when using ExchangedTokenRequestedEvent event

saw-jan:
Allow requesting scope when using `ExchangedTokenRequestedEvent` event.

Usage:
```php
$scopes = ["myscope1", "myscope2"];

$event = new ExchangedTokenRequestedEvent($targetAudience, $scopes);
```


saw-jan:
@julien-nc Could you review it?

julien-nc (commented):
Looks good. It would be nicer to pass the extra scopes as an array that would turned into a string in the end like `implode(' ', $extraScopes)`. Wdyt?

saw-jan:
> Looks good. It would be nicer to pass the extra scopes as an array that would turned into a string in the end like `implode(' ', $extraScopes)`. Wdyt?

That would also be nice. Let me change it to array. Thanks

saw-jan:
> Looks good. It would be nicer to pass the extra scopes as an array that would turned into a string in the end like `implode(' ', $extraScopes)`. Wdyt?

Done!

julien-nc (approved):

saw-jan:
@julien-nc Please merge this if okay to do so. I have no permissions to merge it

PR #1101, opened 2025-04-11
feat: allow requesting scopes when using ExternalTokenRequestedEvent event

saw-jan:
Allow requesting scope when using ExternalTokenRequestedEvent event.

Usage:
```php
$scopes = "myscope1 myscope2";

$event = new ExternalTokenRequestedEvent($scopes);
```

### DIFF
--- a/lib/Service/TokenService.php
+++ b/lib/Service/TokenService.php
@@ -73,7 +73,7 @@ class TokenService {
 	 * @return Token|null Return a token only if it is valid or has been successfully refreshed
 	 * @throws \JsonException
 	 */
-	public function getToken(bool $refreshIfExpired = true): ?Token {
+	public function getToken(bool $refreshIfExpired = true, string $extraScopes = ''): ?Token {
 		$sessionData = $this->session->get(self::SESSION_TOKEN_KEY);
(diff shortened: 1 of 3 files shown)


saw-jan:
@julien-nc Could you review it?

lib/Service/TokenService.php
```diff
@@ -177,6 +181,7 @@ public function refresh(Token $token): Token {
 						'client_secret' => $clientSecret,
 						'grant_type' => 'refresh_token',
 						'refresh_token' => $token->getRefreshToken(),
+						'scope' => $scope,
```
saw-jan, line 184: this not working. maybe we cannot request new scopes while refreshing token :thinking:

saw-jan: We will hit a blockage with `ExternalTokenRequestedEvent` because the login token doesn't contain the extra scope (e.g.: `api_v3`) and we cannot request new scopes while refreshing the token.

julien-nc: It seems expected that the refreshed token has the same scopes.
In this case the scopes must be requested in the provider configuration of user_oidc so that the login token itself already has those scopes.

Issue #1100, opened 2025-04-10
Unable to map nested claim


dragonpil:
Hi there 👋

I'm currently using Nextcloud `30.0.8` with the `user_oidc` app version `7.1.0`. I'm trying to use OpenID Connect for user login and synchronization via EasyVerein as an Identity Provider.

In my configuration, I have added the scopes: openid myself custom
Extra Claims: empty

I have defined for example the following in  the mapping: 
displayName: custom.nickname

However, the displayName field is not populated correctly in Nextcloud. I have increased the log level to 0 and confirmed via the logs what the returned token contains. Here's a sanitized version of the relevant part of the ID token:

```json
{
  "sub": "XXXXXXXX",
  "name": "OpenID Test",
  "username": "XXXX@domain.tld",
  "email": "XXXX@domain.tld",
  "custom": {
    "nickname": "Spitzname"
  },
  ...
}
```
As you can see, the nickname is present in a nested custom claim. According to the [docs](https://github.com/nextcloud/user_oidc), the plugin allows specifying claim paths like custom.nickname, but this does not seem to work in practice (or is unsupported for nested claims?).

Other top-level mappings such as email and sub work perfectly.

❓ Question
Is there currently a supported way to map nested claims (like custom.nickname) to fields such as displayName in Nextcloud?

If not, is this a known limitation or something that could be supported in future releases?

Thanks in advance – and thanks for maintaining this useful plugin 🙌

Best regards,

julien-nc:
Hey, thanks for reporting this issue.

> According to the [docs](https://github.com/nextcloud/user_oidc), the plugin allows specifying claim paths like custom.nickname

Which part of the README are you referring to? I don't think nested attributes are mentioned.

This is indeed a limitation of the current implementation. I will have a look soon to check if there is a simple way to implement support for that.

dragonpil:
Hi,
I re-read everything and couldn't find the position where I originally thought I saw it – sorry for the confusion.

In the meantime, I already started preparing a solution on my side. How is the usual approach here? I'd be happy to prepare a pull request, but I would definitely need help checking whether I’ve changed all relevant places, if the general direction makes sense, and how to properly test it.

I also have a second feature request related to this topic:
It would be great to allow some kind of fallback logic in the attribute mapping – like:

`custom.nickname | profile.name | name |  sub`

… meaning “if the first value does not exist, try the next.”

I already have an idea how to implement this and would include it in the same PR if you think that would be welcome.

[...]

PR #1103, opened 2025-04-11
Issue-1100: Nested & Fallback Claims

dragonpil:
As described in Issue-1100 there is current no possibility to mappe values which are nested in other the the root scope. To enable this we change the way how Mapping-Attributes are used. They should understand the . as object seperator.

Sometimes you like to have a fallback in your mappings. So for displayname for example please take the nickname (if it exists), if it does not exists take the full name .... To support this, we change the way how Mapping-Attributes are used. They should understand the | as an alternative mapping.

WIP: Testing need to be done still fully, I dont know if I changed everything needed but on my side it works.

#1100

### DIFF
--- a/lib/Service/ProvisioningService.php
+++ b/lib/Service/ProvisioningService.php
@@ -59,6 +59,52 @@ class ProvisioningService {
 	}
 
 	/**
+	 * Resolves a claim path like "custom.nickname" or multiple alternatives separated by "|".
+	 * Returns the first found string value, or null if none could be resolved.
+	 */
+	public function getClaimValue(object|array $tokenPayload, string $claimPath, int $providerId): mixed {
+		if ($claimPath === '') {
+			return null;
+		}
+
+		// Check config if dot-notation resolution is enabled
+		$resolveDot = $this->providerService->getSetting($providerId, ProviderService::SETTING_RESOLVE_NESTED_AND_FALLBACK_CLAIMS_MAPPING, '0') === '1';
+
+		if (!$resolveDot) {
+			// fallback to simple access
+			if (is_object($tokenPayload) && property_exists($tokenPayload, $claimPath)) {
+				return $tokenPayload->{$claimPath};
+			} elseif (is_array($tokenPayload) && array_key_exists($claimPath, $tokenPayload)) {
+				return $tokenPayload[$claimPath];
+			}
+			return null;
+		}
+
+		// Support alternatives separated by "|"
+		$alternatives = explode('|', $claimPath);
+
+		foreach ($alternatives as $altPath) {
+			$parts = explode('.', trim($altPath));
+			$value = $tokenPayload;
+
+			foreach ($parts as $part) {
+				if (is_object($value) && property_exists($value, $part)) {
+					$value = $value->{$part};
+				} elseif (is_array($value) && array_key_exists($part, $value)) {
+					$value = $value[$part];
+				} else {
+					continue 2;
+				}
+			}
+
+			if (is_string($value)) {
+				return $value;
+			}
+		}
+
+		return null;
+	}
+	/**
 	 * @param string $tokenUserId
 	 * @param int $providerId
 	 * @param object $idTokenPayload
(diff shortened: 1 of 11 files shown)


julien-nc (changes requested):
That was fast :rocket: 
Looks very good.

* You can run `composer run cs:check` and `composer run cs:fix` to make sure you respect the project code style
* A little bit of documentation in the README would be nice
* The `OCA\UserOIDC\User\Validator\UserInfoValidator` also needs adjustments to get the uid
* Also one question: Is it possible in some cases (depending on the IdP and its configuration) that a claim with a dot does not result in getting nested attributes in the token? Like "custom.plop" would give a token payload like
``` json
{
    "custom.plop": "the value",
}
```
If so, we need to make sure this does not break with this new feature. This would mean making the support for nested attributes (and fallback) optional and disabled by default. What do you think?

dragonpil:
Hi [@julien-nc](https://github.com/julien-nc),

thanks a lot for your super fast and helpful review comments!
I've hopefully addressed everything — and for better readability, I’ve done it in separate commits. I’ll squash them all together once we’re close to merging.

So, in short:

✅ All commits are now signed
✅ Code style is fixed
✅ Missing logic in UserInfoValidator has been added
✅ Your question about dotted claim names: you're absolutely right — depending on the IdP, tokens might include flat keys with dots like "custom.plop". That's why the nested claim resolution is now configurable and disabled by default to avoid breaking existing setups.
I've also added a section to the README explaining the feature and how to enable it via occ.

Let me know what you think — and thanks again!

lib/Service/ProvisioningService.php
```diff
@@ -58,6 +58,54 @@ public function hasOidcUserProvisitioned(string $userId): bool {
...
+	 * Returns the first found string value, or null if none could be resolved.
+	 */
+	public function getClaimValue(object|array $tokenPayload, string $claimPath, int $providerId): ?string {
+		if ($claimPath === '') {
+			return null;
+		}
+
+		// Check config if dot-notation resolution is enabled
+		$resolveDot = false;
+		$resolveDot = $this->providerService->getSetting($providerId, ProviderService::SETTING_RESOLVE_NESTED_AND_FALLBACK_CLAIMS_MAPPING, false);
```
julien-nc, line 72:
```suggestion
		$resolveDot = $this->providerService->getSetting($providerId, ProviderService::SETTING_RESOLVE_NESTED_AND_FALLBACK_CLAIMS_MAPPING, '0') === '1';
```

julien-nc (changes requested):
Awesome, thanks for the changes. As this new `nestedAndFallbackClaims` setting is provider-specific and quite important, it deserves to be set in the UI in my opinion. If you allowed the repo's owners to push in your branch i can add that.

lib/Service/ProvisioningService.php
julien-nc: I took the freedom to push in your branch. I hope that's fine. So the `getSetting` call is fixed.

julien-nc:
So we have one remaining issue with these changes. It was well spotted by Psalm :blue_heart: .
We historically handle the case where the address is a JSON object in the token. Now that `getClaimValue` returns `?string` we can't get an object anymore in the mapped address token attribute.
In Keycloak, for example, it's possible to produce JSON attributes:
image
What I mean is: Independently of how we parse the token payload, we might end up with a value that is something else than a string. Specifically for the `address` attr.

In short, I think we need to adjust the return type of `getClaimValue` to mixed because we don't know which type we return there, an attribute can as well be an object.

I pushed it. Hope you don't mind.

dragonpil:
> Awesome, thanks for the changes. As this new `nestedAndFallbackClaims` setting is provider-specific and quite important, it deserves to be set in the UI in my opinion. If you allowed the repo's owners to push in your branch i can add that.

I just added you as contributor on my fork. Can you help me, how to allow the repo's owner to push in my branch?

julien-nc:
> I just added you as contributor on my fork. Can you help me, how to allow the repo's owner to push in my branch?

It's something you chose when creating the PR. I have the right to push in this PR's branch already. I did push a few commits.

julien-nc:
You can squash. I'll read it all again and merge.

julien-nc:
@dragonpil I realized that the setting was not checked in `LoginController::login` to compute the claim list that is sent in the authorization URL as GET param. Can you review the commit I pushed?

dragonpil:
> @dragonpil I realized that the setting was not checked in `LoginController::login` to compute the claim list that is sent in the authorization URL as GET param. Can you review the commit I pushed?

Hi [julien-nc](https://github.com/julien-nc),
sorry for the delay. yes that's true... but in your commit I just see a change in the readme.md

julien-nc:
Hmmm you squashed all your commits in one, then i pushed one on top which only contains a change in the LoginController.
https://github.com/nextcloud/user_oidc/pull/1103/commits/72249043874e92b7039060af9df993bb7ac238d8

dragonpil:
> Hmmm you squashed all your commits in one, then i pushed one on top which only contains a change in the LoginController. [7224904](https://github.com/nextcloud/user_oidc/commit/72249043874e92b7039060af9df993bb7ac238d8)

ah ok... sorry used the compare button and just saw this: 
https://github.com/nextcloud/user_oidc/compare/5af62e057760a43d94237b4ab854977af0f09825..63814bbd47c73835131ccd55f7cba8fe3f56cafe

so yes I reviewed your change and its fine.

julien-nc (approved):
Thanks again!

PR #1107, opened 2025-04-16
Make settings form footer sticky

julien-nc:
To avoid having to scroll down to find the ok/cancel buttons, the footer is now sticky at the bottom.

image


janepie (approved):
Looks good!

PR #1122, opened 2025-05-20
chore(tests): Cleanup bootstrap.php to be forward-compatible

come-nc:
This uses the new tests/autoload.php from nextcloud/server instead of messing directly with OC private autoloader.

See https://github.com/nextcloud/server/pull/52951 and https://github.com/nextcloud/server/pull/52945 for context.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -26,9 +26,6 @@ jobs:
         include:
             - php-versions: 8.1
               databases: mysql
-              server-versions: stable28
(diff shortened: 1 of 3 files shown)


come-nc:
@julien-nc is stable28 testing still needed?
tests/autoload.php was backported until stable29 but no further.
If really needed I can backport it to stable28 but as it’s EOL I think dropping the test is easier?

julien-nc:
@come-nc Yep, stable28 testing can be dropped.

PR #1123, opened 2025-05-20
Set the gss session data in the controller rather than in the service

julien-nc:
I don't even know how it worked when setting it in the provisioning service.

The logic is now to return the user data from  `ProvisioningService::provisionUser` and set the session value (that will be used by gss master) in the controller.

### DIFF
--- a/lib/Service/ProvisioningService.php
+++ b/lib/Service/ProvisioningService.php
@@ -13,6 +13,7 @@ use OCA\UserOIDC\AppInfo\Application;
 use OCA\UserOIDC\Db\UserMapper;
 use OCA\UserOIDC\Event\AttributeMappedEvent;
 use OCP\Accounts\IAccountManager;
+use OCP\Accounts\PropertyDoesNotExistException;
 use OCP\AppFramework\Db\DoesNotExistException;
 use OCP\AppFramework\Db\MultipleObjectsReturnedException;
 use OCP\DB\Exception;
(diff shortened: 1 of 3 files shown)


julien-nc:
cc @maximelehericy

DaphneMuller:
@juliusknorr  any chance you can give this a review?

lib/Service/ProvisioningService.php
```diff
@@ -104,15 +106,18 @@ public function getClaimValue(object|array $tokenPayload, string $claimPath, int
...
 	 * @param object $idTokenPayload
 	 * @param IUser|null $existingLocalUser
-	 * @return IUser|null
+	 * @return array
 	 * @throws Exception
+	 * @throws PropertyDoesNotExistException
+	 * @throws PreConditionNotMetException
 	 */
-	public function provisionUser(string $tokenUserId, int $providerId, object $idTokenPayload, ?IUser $existingLocalUser = null): ?IUser {
+	public function provisionUser(string $tokenUserId, int $providerId, object $idTokenPayload, ?IUser $existingLocalUser = null): array {
```
juliusknorr, line 120: Not a fan of using arrays that way, but can we annotate the type of the array so psalm can analyze it at least?

julien-nc: Thanks, done. I can't predict what's gonna be in `userData` so I keep a simple `array` type. All good?

> Not a fan of using arrays that way

I agree it's not great. The alternative would be to pass the ISession object from the LoginController to `provisioningService->provisionUser()` to be able to change the session in there. Not elegant either.

julien-nc: I can also create a `ProvisioningResult` class that contains the user and the raw user data. It's a bit cleaner. Wdyt?

julien-nc:
@juliusknorr By the way, any idea why it does not work anymore to store things in the session with an ISession object injected in a service? I'm pretty sure it worked before. Anything changed in the server?

PR #1149, opened 2025-06-28
Support nested claim mapping for groups attribute

### LINKED ISSUE #799: Support for group mapping from mulitple (nested?) entries
We use keycloak and this plugin. We have both realm roles, and nextcloud specific client roles (like `admin` which represents the nextloud admin group). 

Keycloak by default delivers the users realm and client roles like this

```
  "realm_access": {
    "roles": [
      "offline_access",
      "uma_authorization",
      "medical-practitioner",
      "medical-user",
      "user",
      "mpa"
    ]
  },
  "resource_access": {
    "nextcloud": {
      "roles": [
        "admin"
      ]
    }
  }
``` 

So in this case, the groups to apply for nextcloud would be: `[ "offline_access", "uma_authorization", "medical-practitioner", "medical-user", "user", "mpa"]`AND `["admin"]`. At the moment it is only possible to set a single attribute as the source for groups and I am not even sure, if a nested attribute would be accepted.

* So to map only the realm roles, I would have to be able to set `realm_access.roles` as source for `groups` (not sure if this is supported)

But my real feature request is to support a group attribute entry like this `realm_access.roles,resource_access.nextcloud.roles` - this way I would not have to introduce an additional mapper, but nextcloud (with this plugin) could directly use the groups as expected.

andreblanke:
The provider option "Enable nested and fallback claim mappings" recently introduced in #1103 currently doesn't work for the groups attribute mapping since `getSyncGroupsOfToken` uses `$groupsAttribute` for directly accessing properties inside `$idTokenPayload` (instead of using `getClaimValue` to allow for dot-notation resolution).

This is unfortunate because some identity providers, such as Keycloak, nest user roles inside the ID token (e.g., under `realm_access.roles` for Keycloak realm-level and `resource_access.${client_id}.roles` for Keycloak client-level roles).

My PR fixes the issue by adding the method `getClaimValues` and using it inside of `getSyncGroupsOfToken`. `getClaimValues` includes most of the logic of `getClaimValue`, but allows returning non-string values. The original behavior of `getClaimValue` is retained.

Closes #799.

### DIFF
--- a/lib/Service/ProvisioningService.php
+++ b/lib/Service/ProvisioningService.php
@@ -62,9 +62,9 @@ class ProvisioningService {
 
 	/**
 	 * Resolves a claim path like "custom.nickname" or multiple alternatives separated by "|".
-	 * Returns the first found string value, or null if none could be resolved.
+	 * Returns the first found value, or null if none could be resolved.
 	 */
-	public function getClaimValue(object|array $tokenPayload, string $claimPath, int $providerId): mixed {
+	public function getClaimValues(object|array $tokenPayload, string $claimPath, int $providerId): mixed {
 		if ($claimPath === '') {
 			return null;
 		}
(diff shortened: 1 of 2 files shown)


julien-nc:
To fix the tests, you need to extend what the mocked `getSetting` method will return for `SETTING_RESOLVE_NESTED_AND_FALLBACK_CLAIMS_MAPPING`.

I'll check if I can push in your branch and maybe do it.

julien-nc:
Fixed the tests, rebased on main, all good.

andreblanke:
Thank you, too, for the merge and the test fix.

PR #1151, opened 2025-07-05
feat: allow self-signed SSL verification and support for oidc prompt

elyerr:
### Summary

This PR introduces two key improvements to enhance development and integration flexibility with custom OAuth2/OIDC providers:
 - Features Added
    - Self-signed certificate support:   A new system config flag httpclient.allowselfsigned enables accepting self-signed SSL certificates during development or internal deployments.    When enabled, SSL verification is disabled (verify => false) in HTTP client requests.

- Custom OIDC prompt support:
   - Adds support for the standard OIDC prompt values: none, login, and consent.
        Introduces a custom internal prompt designed specifically for private applications using [OAuth2 Passport Server](https://github.com/elyerr/oauth2-passport-server).

        Documentation for this prompt is available here: [Prompts Supported](https://gitlab.com/elyerr/oauth2-passport-server/-/wikis/home/prompts-supported)

##  Note
Use of self-signed certificates is intended only for development or trusted internal networks. It should be disabled in production environments to ensure secure communication.

### DIFF
--- a/lib/Helper/HttpClientHelper.php
+++ b/lib/Helper/HttpClientHelper.php
@@ -14,8 +14,7 @@ require_once __DIR__ . '/../../vendor/autoload.php';
 use Id4me\RP\HttpClient;
 use OCP\Http\Client\IClientService;
 
-class HttpClientHelper implements HttpClient
-{
+class HttpClientHelper implements HttpClient {
 
 	public function __construct(
 		private IClientService $clientService,
@@ -23,8 +22,7 @@ class HttpClientHelper implements HttpClient
 	) {
 	}
 
-	public function get($url, array $options = [])
-	{
+	public function get($url, array $options = []) {
 		$client = $this->clientService->newClient();
 
 		if ($this->config->getSystemValue('httpclient.allowselfsigned', false)) {
@@ -34,8 +32,7 @@ class HttpClientHelper implements HttpClient
 		return $client->get($url, $options)->getBody();
 	}
 
-	public function post($url, $body, array $headers = [])
-	{
+	public function post($url, $body, array $headers = []) {
 		$client = $this->clientService->newClient();
 
 		$options = [
--- a/lib/Helper/HttpClientHelper.php
+++ b/lib/Helper/HttpClientHelper.php
@@ -22,7 +22,7 @@ class HttpClientHelper implements HttpClient {
 	) {
 	}
 
-	public function get($url, array $options = []) {
+	public function get($url, array $headers = [], array $options = []) {
 		$client = $this->clientService->newClient();
 
 		if ($this->config->getSystemValue('httpclient.allowselfsigned', false)) {
--- a/lib/Helper/HttpClientHelper.php
+++ b/lib/Helper/HttpClientHelper.php
@@ -8,32 +8,55 @@ declare(strict_types=1);
 
 namespace OCA\UserOIDC\Helper;
 
-use OCP\Http\Client\IClientService;
+use OCP\IConfig;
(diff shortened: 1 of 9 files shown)


lib/Helper/HttpClientHelper.php
```diff
@@ -8,32 +8,55 @@
...
-		return $client->get($url, [
-			'headers' => $headers,
-		])->getBody();
+		if ($this->config->getSystemValue('httpclient.allowselfsigned', false)) {
+			$options = array_merge(
+				$options,
+				[
+					'verify' => false
+				]
+			);
```
julien-nc, line 36:
```suggestion
			$options['verify'] = false;
```

lib/Helper/HttpClientHelper.php
```diff
@@ -8,32 +8,55 @@
...
-		])->getBody();
+		];
+
+		if ($this->config->getSystemValue('httpclient.allowselfsigned', false)) {
+			$options = array_merge(
+				$options,
+				[
+					'verify' => false
+				]
+			);
```
julien-nc, line 57:
```suggestion
			$options['verify'] = false;
```

julien-nc (changes requested):
Nice, thank you. Can you remove all the formatting changes and make sure your additions use the same code style?
You can run `composer run cs:fix` if you want.
I will try this out and give another review.

julien-nc:
Works fine with Keycloak. I checked with prompt set to none, login and consent (with `Consent required` enabled on the IdP side).

Not sure how the user is supposed to log in when prompt is none. It seems like the IdP expects the credentials in the authorization request because it returns `login_required` when the flow is initiated by user_oidc.

Maybe you should add an extra detail in the README: For the consent page to be displayed with Keycloak, `prompt=consent` must be set in user_oidc AND `Consent required` must be enable on the Keycloak side.

DaphneMuller:
@elyerr hi! would you be up for doing the changes or would you prefer if we finish the pr?

julien-nc:
@elyerr Thanks for the adjustment.
I rebased on main, ran `composer run cs:fix` and adjusted the README.

One last thing: the `httpclient.allowselfsigned` option only affects user_oidc so it makes sense if it is inside the `user_oidc` array in config.php. Could make that change?

julien-nc (approved):

YeapGuy:
This broke my Nextcloud instance's intended behavior. What am I supposed to do here when I just want it to act the way it did before this PR??
Whatever I set in the config, it goes with prompt=consent in the Authentik URL, and my users have to approve login every single time. And I feel like none of the options would work for me anyway based on their descriptions in the attached external docs. Authentik doesn't support the non-standard 'internal' option iirc, and I've no idea how I would emulate that behavior now (‒ it acted like that by itself before this PR)

elyerr:
@YeapGuy, In OAuth2 and OpenID Connect, the use of these three prompts (consent, login, none) is a standard.
You can use none in the prompt parameter to make your authorization server skip asking the user for consent, but this requires that the user has already signed in and authorized the application beforehand.
If you want something more customized, you can modify your authorization server and create a new prompt that only works with your server. For example, in my server I have a custom prompt called "internal", which allows the owner’s applications to skip asking for consent even the first time the app is used. If the user is not logged in, it will force them to log in and then redirect them back to the application without asking for consent. However, this should only be applied to 100% trusted applications.
On my server (https://github.com/elyerr/oauth2-passport-server), you can check it out and try it.
Apps are registered in admin → clients, and then you can replicate that setup in your own authorization server.

PR #1157, opened 2025-07-21
Add warning log with more data when there is a state mismatch

julien-nc:
It was a debug log before and it didn't contain data.


janepie (approved):
Nice idea!

PR #1165, opened 2025-07-29
Fix/store login token

elyerr:
Pull Request: Fix OIDC Session Token Storage

This PR fixes an issue where the store_login_token method was not properly storing authentication data in the session. As a result, essential information returned by the authorization server—such as id_token, access_token, and other claims—was not available for later use.

By correcting this behavior, the session now correctly retains the OIDC token data, which can be reliably used for validation, user identification, or other post-login processes that depend on the authorization server's response.

### DIFF
--- a/lib/Controller/LoginController.php
+++ b/lib/Controller/LoginController.php
@@ -591,7 +591,7 @@ class LoginController extends BaseOidcController {
 			$this->eventDispatcher->dispatchTyped(new UserLoggedInEvent($user, $user->getUID(), null, false));
 		}
 
-		$storeLoginTokenEnabled = $this->config->getAppValue(Application::APP_ID, 'store_login_token', '0') === '1';
+		$storeLoginTokenEnabled = in_array($oidcSystemConfig['store_login_token'], ['1', 1, true]);
 		if ($storeLoginTokenEnabled) {
 			// store all token information for potential token exchange requests
 			$tokenData = array_merge(
--- a/lib/Listener/ExternalTokenRequestedListener.php
+++ b/lib/Listener/ExternalTokenRequestedListener.php
@@ -23,12 +23,19 @@ use Psr\Log\LoggerInterface;
  */
 class ExternalTokenRequestedListener implements IEventListener {
 
+	/**
+	 * User oidc config
+	 * @var 
+	 */
+	private $oidcSystemConfig;
+
 	public function __construct(
 		private IUserSession $userSession,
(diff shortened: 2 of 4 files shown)


julien-nc:
Thank you for the PR. Could you put your commits on top of main and avoid changing the code style? I have a hard time reading your changeset :grin:

elyerr:
Sorry @julien-nc , I made the change from a modified branch, I just fixed that  and please accept them so I can update my branch from the main branch.

julien-nc:
Thank you for the changes.

I'm not sure why you are now getting the value from the system config (config.php) because the value is actually stored in the database, in the app config. See https://github.com/nextcloud/user_oidc/blob/aa97baafcdf56e35f6621f3b546a1968443be607/lib/Controller/SettingsController.php#L178-L180

Are you saying that the value is not obtained correctly with 
`$this->config->getAppValue(Application::APP_ID, 'store_login_token', '0') === '1'`
?

elyerr:
@julien-nc , Exactly, when I tried to retrieve the data provided by the authorization server — which should be stored during login when store_login_token = "1" is set with user_oidc — it wasn’t being saved. After reviewing the plugin, I realized that feature wasn’t working properly, so I fixed it. In my case, I need to store that data because my authorization server sends me an access_token that I use to connect back to it and perform some queries with other applications.

julien-nc:
I'm sorry I don't see why it does not work when using `$this->config->getAppValue(Application::APP_ID, 'store_login_token', '0') === '1'`. Are you sure this is not working with the main branch or v7.3.0?

I just checked and it works fine for me in main. In `TokenService::checkLoginToken`, the `$storeLoginTokenEnabled` var has the correct value according to what is set in the admin settings with the image checkbox.
The value is stored correctly in oc_appconfig and retrieved correctly everywhere.

Am I missing something?

elyerr:
@julien-nc,  I’m using version 32.0.0.1 of Nextcloud, and the default configuration of the latest version of user_oidc didn’t work.
```
'user_oidc' => [
    'prompt' => 'internal',
    'store_login_token' => '1' // '1',1,true
  ],
```

But after making some corrections, it worked normally.

julien-nc:
The `store_login_token` value is not supposed to be stored in config.php. It is stored in oc_appconfig. The checkbox in the admin settings has an effect on oc_appconfig, not config.php. I still don't get why you switched to config.php.

PR #1170, opened 2025-08-07
Fix soft auto provisioning

### LINKED ISSUE #1169: soft_auto_provision=false is broken when unique user IDs are disabled
When auto_provision = true, and soft_auto_provision = false, we're supposed to check if the user is already present in another backend [here](https://github.com/nextcloud/user_oidc/blob/aa97baafcdf56e35f6621f3b546a1968443be607/lib/Controller/LoginController.php#L547).

When at the same time, "Use unique user IDs" is disabled (so that the "sub" from the provider is used directly as the NC user id), this check appears to break though: A new user gets provisioned nicely on the first login via OIDC, but on the second time, login fails with "User conflict" [here](https://github.com/nextcloud/user_oidc/blob/aa97baafcdf56e35f6621f3b546a1968443be607/lib/Controller/LoginController.php#L550).

Possibly `userManager->get($tokenUserId);` returns the OIDC user now, because there's a match between "sub" from the identity provider and "user_id" in the NC tables. In that case I guess the solution would be to check the backend of the returned user object and ignore if the backend is ourself?

julien-nc:
closes #1169

Prevent user conflict when soft-auto-provisioning is on and the user exists but is a user_oidc one.

cc @jonas2515

### DIFF
--- a/lib/Controller/LoginController.php
+++ b/lib/Controller/LoginController.php
@@ -538,25 +538,25 @@ class LoginController extends BaseOidcController {
 			$this->ldapService->syncUser($userId);
 		}
 
-		$userFromOtherBackend = $this->userManager->get($userId);
-		if ($userFromOtherBackend !== null && $this->ldapService->isLdapDeletedUser($userFromOtherBackend)) {
-			$userFromOtherBackend = null;
+		$existingUser = $this->userManager->get($userId);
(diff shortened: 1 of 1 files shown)


jonas2515:
Yay, thanks for the quick fix! Can confirm it works as expected now!

julien-nc:
Thank you for the feedback. A release is on its way.

PR #1181, opened 2025-08-20
Handle session token revocation

### LINKED ISSUE #1159: Session revoking logic
To continue the discussion from https://github.com/nextcloud/server/issues/53942

julien-nc:
:warning: This depends on https://github.com/nextcloud/server/pull/54545 which will be included in Nextcloud 32.

When a user logs in twice (with user_oidc) and invalidates the other session in the security settings, for example, it does not kill the IdP session so, in the browser session that is revoked, it is possible to complete the login flow without authenticating again in the IdP.

This PR adds a listener on a new event dispatched when an auth token is invalidated: `OCP\Authentication\Events\TokenInvalidatedEvent`
If we have an Oidc session that matches this invalidated token, we call the end_session_endpoint to make sure the related IdP session is terminated.

We also cleanup the session entry (in a user_oidc table) when this happens.

New columns in the `oc_user_oidc_session` table:
* `user_id` is used with `authtoken_id` to get the session when a token is invalidated
* `provider_id` is necessary to get the `client_id` so we can send it as a param to the end_session_endpoint
* `id_token` is the login ID token, we also need it as a param to the end_session_endpoint
* `idp_session_closed` helps with backchannel logout (explained below)

To make sure we don't mess with the backchannel logout, there is a new `idp_session_closed` column in the `oc_user_oidc_session` table. When we receive a backchannel logout request, we know that the IdP is currently killing one of its sessions. We call `authTokenProvider->invalidateTokenById` to kill the related Nextcloud session. So we will receive the `OCP\Authentication\Events\TokenInvalidatedEvent` but this time we don't need to make a request to the end_session_endpoint, we know the IdP session is already terminated. So we store that information in the session in the `idp_session_closed` column so the listener can ignore the event related with this Oidc session.

closes #1159
closes https://github.com/nextcloud/server/issues/53942

### DIFF
--- a/lib/Db/Session.php
+++ b/lib/Db/Session.php
@@ -11,38 +11,49 @@ namespace OCA\UserOIDC\Db;
 use OCP\AppFramework\Db\Entity;
 
 /**
- * @method string getSid()
- * @method void setSid(string $sid)
- * @method string getSub()
- * @method void setSub(string $sub)
- * @method string getIss()
- * @method void setIss(string $iss)
- * @method int getAuthtokenId()
- * @method void setAuthtokenId(int $authtokenId)
- * @method string getNcSessionId()
- * @method void setNcSessionId(string $ncSessionId)
- * @method int getCreatedAt()
- * @method void setCreatedAt(int $createdAt)
+ * @method \string getSid()
+ * @method \void setSid(string $sid)
+ * @method \string getSub()
+ * @method \void setSub(string $sub)
+ * @method \string getIss()
+ * @method \void setIss(string $iss)
+ * @method \int getAuthtokenId()
+ * @method \void setAuthtokenId(int $authtokenId)
+ * @method \string getNcSessionId()
+ * @method \void setNcSessionId(string $ncSessionId)
+ * @method \int getCreatedAt()
+ * @method \void setCreatedAt(int $createdAt)
+ * @method \string|\null getIdToken()
+ * @method \void setIdToken(?string $idToken)
+ * @method \string|\null getUserId()
+ * @method \void setUserId(?string $userId)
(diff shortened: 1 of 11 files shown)


lib/Listener/TokenInvalidatedListener.php
```diff
@@ -0,0 +1,130 @@
...
+		} catch (Exception|DoesNotExistException|MultipleObjectsReturnedException $e) {
+			$this->logger->warning('[TokenInvalidatedListener] Could not find the OIDC session related with an invalidated token', [
+				'token_id' => $eventTokenId,
+				'user_id' => $eventTokenUserId,
+				'exception' => $e,
+			]);
+			return;
+		}
+		// we have nothing to do if we know the idp session is already closed
+		if ($oidcSession->getIdpSessionClosed() !== 0) {
```
janepie, line 67: Why not `==1`? Can it be `null` for some reason?

lib/Db/Session.php
```diff
@@ -51,18 +62,25 @@ public function __construct() {
...
+			'id' => $this->getId(),
+			'sid' => $this->getSid(),
+			'sub' => $this->getSub(),
+			'iss' => $this->getIss(),
+			'authtoken_id' => $this->getAuthtokenId(),
+			'nc_session_id' => $this->getNcSessionId(),
+			'created_at' => $this->getCreatedAt(),
+			'user_id' => $this->getUserId(),
+			'provider_id' => $this->getProviderId(),
+			'idp_session_closed' => $this->getIdpSessionClosed() !== 0,
```
janepie, line 83: same here

janepie (approved):
Nitpicks, looks good overall!

Issue #836, opened 2024-04-15
Does the user_oidc application support notes and other non-OCS API endpoints?


akhil1508:
- I am trying to figure out the best OIDC plugin to use for my nextcloud setup
- For non-OCS API endpoints, I need to know if this plugin supports them or whether we get an error from `CORSMiddleware` like `CORS requires BASIC auth` 
  - From https://github.com/nextcloud/server/blob/10fc78a9ea0d8a2081c1185f8f4c4b63b631d88e/lib/private/AppFramework/Middleware/Security/CORSMiddleware.php#L102 
  - As notes and notes mobile client(login with SSO) is a huge  use-case for my setup.. 

- P.S. I'm not sure if this is the right place for a question like this or if I should be asking in the forum; forum seems less developer oriented to me :thinking:. Apologies in advance if this is the wrong place

PR #1221, opened 2025-10-12
Fix Bearer token authentication for non-OCS endpoints

cbcoutinho:
When using Bearer token authentication with OIDC, API requests to endpoints with `@CORS` annotations (like Notes API) were failing with 401 Unauthorized errors. This occurred because:

1. Bearer token validation successfully authenticated the user
2. A session was created for the authenticated user
3. Nextcloud's CORSMiddleware detected the logged-in session but no
   CSRF token, causing it to call session->logout()
4. The logout invalidated the session, breaking the API request

This fix sets the 'app_api' session flag during Bearer token authentication, which instructs CORSMiddleware to skip the CSRF check and logout logic. This is the same mechanism used by Nextcloud's AppAPI framework for external application authentication.

The flag is set at all successful Bearer token authentication points:
- Line 243: After OIDC Identity Provider validation
- Line 310: After auto-provisioning with bearer provisioning
- Line 315: After existing user authentication
- Line 337: After LDAP user sync

---

Fixes: #836 regarding enabling OIDC for non-OCS endpoints like Notes app, etc.


julien-nc:
Thanks for the fix suggestion. We are discussing it.

So the problematic case is an ApiController using the `#[CORS]` method attribute like in https://github.com/nextcloud/notes/blob/main/lib/Controller/NotesApiController.php .

So if i understand correctly, the problem is that even requests (to the Notes app API for example) that are NOT made by a browser (but with CURL for example) will fail. Right @cbcoutinho ?

cbcoutinho:
Hi @julien-nc, yes - that's correct.

In my limited investigation, it appears that any request to a non-OCS endpoint outside of a browser is rejected.

It's my understanding that this is because the `CORSMiddleware.php` class logs out users that do not have a CSRF token - see:
[/lib/private/AppFramework/Middleware/Security/CORSMiddleware.php](https://github.com/nextcloud/server/blob/65f52c47b83a2511e8e18f9c36d8644196187af3/lib/private/AppFramework/Middleware/Security/CORSMiddleware.php#L80-L84)

The `app_api` session flag is the established Nextcloud pattern for API-based authentication. It's used throughout core to distinguish API authentication from browser sessions. Key examples:
- [CORSMiddleware.php:81](https://github.com/nextcloud/server/blob/65f52c47b83a2511e8e18f9c36d8644196187af3/lib/private/AppFramework/Middleware/Security/CORSMiddleware.php#L81) - Skips CSRF check for app_api sessions
- [TwoFactorAuth/Manager.php:304](https://github.com/nextcloud/server/blob/65f52c47b83a2511e8e18f9c36d8644196187af3/lib/private/Authentication/TwoFactorAuth/Manager.php#L304) - Skips 2FA (same as app passwords)
- [SecurityMiddleware.php:121](https://github.com/nextcloud/server/blob/65f52c47b83a2511e8e18f9c36d8644196187af3/lib/private/AppFramework/Middleware/Security/SecurityMiddleware.php#L121) - Used for ExApp authentication

This ensures Bearer token authentication works with `@CORS` endpoints (like Notes API) without triggering browser-specific security checks like CSRF, which don't apply to token-based API requests.

edward-ly:
Hello, FYI @julien-nc is currently on vacation until Oct. 28, so unfortunately, discussions are on hold until then.

Pro tip: in the meantime, for a cleaner commit history, we'd suggest rebasing your branch so that the base is up-to-date with `main` (rather than having multiple empty merge commits):
```sh
git checkout main
git pull
git checkout fix/bearer-auth-cors-session-logout
git rebase main
git push -f
```

cbcoutinho:
Thanks for the update @edward-ly - no rush.

I'll maintain this branch and make sure that it is rebased on `main` until the PR is merged or closed.

oleksandr-nc:
Cons/risks of current solution:

- **Scope of app_api**: it also disables `2FA check`s for the session and remove AppAPI requests from rate limiting. Using it for `OIDC API` calls is functionally consistent with app passwords/external apps, but it is broader than "just skip CORSMiddleware logout"
- **Session semantics**: if browser ends up reusing that session cookie, the session is now an `API session` skipping 2FA. In practice, modern browsers should third‑party cookie use and most API clients do not send browser cookies, but this is a theoretical thing to be aware of (server code specifically checks the session flag, not the presence of cookies)
- **Tight coupling to AppAPI naming**: `app_api` flag was introduced for ExApps. Reusing it for OIDC Bearer is pragmatic, but semantically it is two different concepts.

Solutions I can propose:

1. Extend `CORSMiddleware` to also accept Bearer auth (**without relying on app_api**). Somewhere in `CORSMiddleware::beforeController()` if the route has `#[CORS]` and the request has `Authorization: Bearer ...`, add the skip of basic‑auth fallback/logout. (motivation/inspiration will be how `SecurityMiddleware` already done)

    Server issues about this: https://github.com/nextcloud/server/issues/44365

2. Add route attribute (e.g., `#[TokenAuthAllowed]`) and honor it in `CORSMiddleware`

    Rather than wide **Bearer allowed on all #[CORS]**, introduce explicit attribute that apps can add where Bearer should be accepted without `CSRF/Basic`.
    
3. Create separate flag recognized by `CORSMiddleware`/`SecurityMiddleware` that bypasses `CSRF` for `CORS` routes like `app_api` does, but does not inherit all AppAPI semantics.

4. Change apps (e.g., Notes) to expose OCS wrappers for their APIs (OCS already treats Bearer as valid)


I lean towards point 4 as the most conceptual cleanest one. Except point 4 I like the point 2.

cbcoutinho:
Thanks for the overview @oleksandr-nc, I've been learning a lot about how Nextcloud handles auth via these PRs, and I appreciate your response.

My preference (beyond this PR) would be your first solution - allow `Bearer` tokens to bypass CORS check within the Nextcloud server itself. To facilitate the discussion, I've opened PR https://github.com/nextcloud/server/pull/55878

edward-ly (changes requested):
To be confirmed by @julien-nc when he's back, but I agree with @oleksandr-nc in the meantime that we shouldn't rely on app_api to do this which does (and may in the future) do something different than what is intended here. I don't have a strong preference for any of the 4 solutions as of now.

cbcoutinho:
Friendly ping @julien-nc - welcome back from holidays.

Would you mind reviewing either this PR or the one targeting the nextcloud server itself in https://github.com/nextcloud/server/pull/55878

cbcoutinho:
Howdy @edward-ly @oleksandr-nc 👋

There is still interest in this feature. Has there been a decision made internally on how to proceed with Bearer auth and whether to allow it in the `user_oidc` app or in the server itself?

https://github.com/nextcloud/server/pull/55878

cbcoutinho:
Hi @julien-nc @oleksandr-nc  friendly ping - could you review this PR or related PR in [nextcloud/server#55878](https://github.com/nextcloud/server/pull/55878)?

ykorzikowski:
Hi @cbcoutinho ,
I just ran into the same issue getting 403 with the apps API. Is there any progress on this issue? Is there any workaround?

cbcoutinho:
Hi @ykorzikowski - unfortunately bearer tokens are currently not accepted by Nextcloud.

PR #1254, opened 2025-11-15
fix: Update EdDSA mapping to OKP in DiscoveryService

joshtrichards:
RFC 8037 defines EdDSA public keys as `kty='OKP'`. Current master derives expected `kty` from `alg`; with the wrong mapping it filters out valid OKP keys and fails with “No matching key found.”

#1236 almost fixed it, but this should make it work fully now.


solracsf (approved):

julien-nc (approved):
Thank you!

PR #1268, opened 2025-12-09
Feature: Add configurable protected groups setting

o-m-n-i:
Add configurable protected groups setting

- Add SETTING_PROTECTED_GROUPS to allow admins to configure which groups
  should never have users removed from them
- Default protected groups are 'users' and 'admin'
- Add UI field in admin settings to configure protected groups per provider
- Update group whitelist regex help text to show example with multiple patterns
- Add translation strings for protected groups setting


o-m-n-i:
Wrong repo!

Issue #1175, opened 2025-08-17
Race condition in TokenService causes logouts on token refresh


albgus:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

### Steps to reproduce
1. Configure OIDC with refresh tokens to extend the session. Set a short access token validity to trigger the issue faster 
2. Login with OIDC and leave a tab open with the Nextcloud web UI
3. After some time of inactivity return to the tab and try to browse around.

### Expected behaviour
The session should continue uninterrupted

### Actual behaviour
After the access token is expired, any access of Nextcloud in an already open tab will cause a logout event.

Diggin into the logs it appears to be because a request is sent to `/ocs/v2.php/apps/notifications/api/v2/notifications` and slightly later `/ocs/v2.php/apps/user_status/api/v1/heartbeat?format=json`. This causes both of these to run in parallel, and the first request will use the stored refresh token to refresh the session, then the second request will attempt to refresh the session. This causes a failure since it appears to use the refresh token that was read before the other request used it, so it's now expired.

Not sure what the best solution here is but it would seem that either some database locking would be needed to ensure that the token is not modified by parallel requests.

<details>
<summary>Nextcloud log</summary>

```
2025-08-17 00:22:35.577 user=-- url=/ocs/v2.php/apps/user_status/api/v1/heartbeat?format=json message=[TokenService] reauthenticate
2025-08-17 00:22:35.577 user=albgus url=/ocs/v2.php/apps/user_status/api/v1/heartbeat?format=json message=[TokenService] checkLoginToken: token is still expired -> reauthenticate
2025-08-17 00:22:35.577 user=albgus url=/ocs/v2.php/apps/user_status/api/v1/heartbeat?format=json message=[TokenService] Failed to refresh token
2025-08-17 00:22:35.577 user=albgus url=/ocs/v2.php/apps/notifications/api/v2/notifications message=[TokenService] Store token in the session
2025-08-17 00:22:35.577 user=albgus url=/ocs/v2.php/apps/notifications/api/v2/notifications message=[TokenService] ---- Refresh token success
[...]

PR #1178, opened 2025-08-18
Prevent multiple simultaneous token refresh

julien-nc:
Store a lock in the php session.

closes #1175

PR #1277, opened 2025-12-18
fix(token): Implement locking for token refresh

solracsf:
Fix https://github.com/nextcloud/user_oidc/issues/1175

Added locking mechanism to prevent race conditions during token refresh.
Enhanced error handling and logging for token refresh process.

The fix uses `ILockingProvider` to ensure atomic token refresh:​

1. Per-session locking (prevents race within same user's requests)
2. First request acquires lock and refreshes token successfully, or,
3. a second request hits lock and waits 100ms, retries in loop (for 5 seconds max), then reads the already-refreshed token from session

### DIFF
--- a/lib/Service/TokenService.php
+++ b/lib/Service/TokenService.php
@@ -32,6 +32,8 @@ use OCP\IRequest;
 use OCP\ISession;
 use OCP\IURLGenerator;
 use OCP\IUserSession;
+use OCP\Lock\ILockingProvider;
+use OCP\Lock\LockedException;
 use OCP\PreConditionNotMetException;
 use OCP\Security\ICrypto;
 use OCP\Session\Exceptions\SessionNotAvailableException;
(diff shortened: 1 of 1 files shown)


julien-nc (commented):
Thank you very much. Code looks good. A few questions then I'll make some tests and we're good to go.

solracsf:
You're right. The token refresh involves network requests to the OAuth2 provider, which typically takes 200-500ms or more. The 100ms wait was pointless because Process A will still be in the middle of the HTTP call...

I've pushed a retry mechanism to accomplish this :

Process A:
- Gets lock immediately
- Refreshes token (network request takes 300ms)
- Stores new token in session
- Releases lock

Process B:
- Tries to get lock, but get `LockedException`
- Waits 100ms, retries, waits, retries, acquires lock (A released it)
- Double-checks session, but Token already fresh
- Returns cached token without making redundant HTTP request
- Releases lock

This (should) prevent both the race condition and unnecessary duplicate token refreshes.

solracsf:
I've just found out that this is a bit of a duplicate of https://github.com/nextcloud/user_oidc/pull/1178 but with another approach... sorry! 🙈

julien-nc:
Please don't be sorry. I'm really not sure writing a lock in the Php session is a viable solution because I don't exactly know when Symfony reads and writes the session data. So we might get outdated data when reading the lock. Using the lock mechanism like you did + the session ID as identifier seems like a better approach.

PR #1286, opened 2026-01-12
More checks when getting an avatar from a URL

julien-nc:
More checks when getting an avatar from a URL.
We only accept image content type payloads when retrieving an avatar with a URL.

### DIFF
--- a/lib/Service/ProvisioningService.php
+++ b/lib/Service/ProvisioningService.php
@@ -446,7 +446,13 @@ class ProvisioningService {
 		if (filter_var($avatarAttribute, FILTER_VALIDATE_URL)) {
 			$client = $this->clientService->newClient();
 			try {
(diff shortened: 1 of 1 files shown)


rhclayto:
This doesn't work for me. `$contentType = $response->getHeader('Content-Type')[0] ?? '';` is suppsoed to return an array (as far as I can tell from https://www.php-fig.org/psr/psr-7/ ), but it seems to be returning a string! For me it is any way, & I don't know why. So $contentType ends up being just the first character of the stong, so for instance, 'image/jpeg' becomes 'i'. And then this check is false `if (!in_array($contentType, ['image/jpeg', 'image/png', 'image/gif'], true)) {`. 

I changed it to `$contentType = $response->getHeader('Content-Type') ?? '';`  & it's working perfectly. Any idea why?

Issue #561, opened 2023-01-13
Figure out a proper way to trigger relevant LoginChain commands


juliushaertl:
There might be quite a few things from https://github.com/nextcloud/server/blob/215aef3cbdc1963be1bb6bca5218ee0a4b7f1665/lib/private/Authentication/Login/Chain.php#L79-L91 that we don't run during our LoginController when a user logs in. Might be good to have a separate LoginChain to still perform those tasks.

PR #1333, opened 2026-02-17
fix(jwks): filter unsupported key types in fixJwksAlg

### LINKED ISSUE #823: DomainException: Unrecognised or unsupported EC curve at apps/user_oidc/lib/Vendor/Firebase/JWT/JWK.php line 145

strobelpierre:
## Summary

Fixes #823 — `JWK::parseKeySet()` crashes with `DomainException: Unrecognised or unsupported EC curve` when the OIDC provider's JWKS endpoint returns keys with types that Firebase JWT cannot parse (e.g. EC P-521 curve, certain OKP subtypes).

### Root cause

`fixJwksAlg()` already identifies the matching key by comparing `kty` (key type family), but it returns the **entire JWKS** — including keys with incompatible types. Firebase JWT's `parseKeySet()` then iterates all keys and throws on the first unsupported one instead of skipping it ([firebase/php-jwt#561](https://github.com/firebase/php-jwt/issues/561)).

### Fix

Filter the JWKS to only include keys matching the expected `kty` before returning. This is a 5-line addition after the existing matching logic:

```php
$jwks['keys'] = array_values(array_filter($jwks['keys'], function ($key) use ($expectedKty) {
    return ($key['kty'] ?? null) === $expectedKty;
}));
```

This way `parseKeySet()` never encounters unsupported key types.

### Why not just update Firebase JWT?

Firebase JWT v6 and v7 both have this behavior — `parseKey()` throws `DomainException` for unsupported curves instead of returning `null`. The upstream issue (#561) has been open since 2023 with no fix merged.

### Supersedes

This PR supersedes the draft PR #868 which only hardcoded P-521 removal. This approach is more general (filters by `kty` family) and works against the current `main` branch.

## Test plan

- [x] Added unit tests for `fixJwksAlg` with mixed key types (RSA + P-521 EC + OKP)
- [x] Added unit test verifying EC keys are preserved when JWT uses ES256

### DIFF
--- a/lib/Service/DiscoveryService.php
+++ b/lib/Service/DiscoveryService.php
@@ -196,16 +196,18 @@ class DiscoveryService {
 			throw new \RuntimeException('Invalid JWKS: missing "keys" array');
 		}
 
+		// Filter out keys with incompatible key types to prevent
+		// Firebase JWT from failing on unsupported curves (e.g. P-521, Ed448).
+		// See https://github.com/firebase/php-jwt/issues/561
+		$jwks['keys'] = array_values(array_filter($jwks['keys'], function ($key) use ($expectedKty) {
+			return ($key['kty'] ?? null) === $expectedKty;
+		}));
+		$keys = $jwks['keys'];
+
 		$matchingIndex = null;
 
 		foreach ($keys as $index => $key) {
-			$keyKty = $key['kty'] ?? null;
 			$keyUse = $key['use'] ?? null;
-
-			// Skip keys with incompatible type
-			if ($keyKty !== $expectedKty) {
-				continue;
-			}
 
 			// Skip keys not intended for signature
 			if ($keyUse !== null && $keyUse !== 'sig') {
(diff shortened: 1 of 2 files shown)


strobelpierre:
> Thanks a lot for this PR!
> 
> Why not filtering before the foreach loop? We skip the keys with the same condition as your filter anyway.

@julien-nc , done :D thx for the idea !

strobelpierre:
Hi @julien-nc 👋

I've rebased the branch on latest `main` and squashed into a single clean commit. The CI workflows are pending approval — could you (or @juliusknorr / @rullzer) approve the workflow runs so the required checks can pass?

Thanks!

geoffspace:
Tested and works for us!

strobelpierre:
@geoffspace , thx for testing ❤️ 
@julien-nc  , thx for merging 😄

PR #1336, opened 2026-02-17
Canonicalize locale claim

alejo7797:
[The OpenID Connect spec](https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims) has the value of the `locale` claim formatted as a [BCP 47](https://www.rfc-editor.org/info/bcp47) language tag, e.g. `en-US` or `fr-CA`.

Nextcloud [expects](https://github.com/nextcloud/server/blob/v32.0.6/resources/locales.json) the locale in ICU format, e.g. `en_US` or `fr_CA`.

This change converts between the two formats during the user provisioning step.

### DIFF
--- a/lib/Service/ProvisioningService.php
+++ b/lib/Service/ProvisioningService.php
@@ -8,6 +8,7 @@
 namespace OCA\UserOIDC\Service;
 
 use InvalidArgumentException;
+use Locale;
 use OC\Accounts\AccountManager;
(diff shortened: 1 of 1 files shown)


julien-nc (approved):
Thank you very much!
Tested and approved.
`Locale::canonicalize` leaves whatever it does not know untouched.
I hope you don't mind, I added a comment.

alejo7797:
Looks great, thank you!

Issue #1121, opened 2025-05-17
Google OIDC login fails in `user_oidc v7.2.0` due to unconditionally injected `claims`


vicentalonso:
❌ Google OIDC login fails in `user_oidc v7.2.0` due to unconditionally injected `claims`

## 🧩 Affected system
- Nextcloud Hub 10 (version 31.0.4.1)
- App: user_oidc v7.2.0
- OIDC Provider: Google (OpenID Connect)
- Server: Debian 12, installed manually

## ❗ Symptom
When trying to log in with Google, the following error is returned by Google:

```
Error 400: invalid_request  
Invalid parameter value for claims: Invalid claims: claims must have a single object value "id_token"
```

## 🔍 Cause
Even with the following in `config.php`:
```php
'enable_default_claims' => false,
```

the app still sends this block:

```json
"claims": {
  "id_token": { "email": null, "name": null, ... },
  "userinfo": { "email": null, "name": null, ... }
}
```

This breaks compatibility with Google, which does not allow unstructured or unnecessary `claims` blocks.

## ✅ Workaround
To make login work, I had to manually edit:

**File:**  
`apps/user_oidc/lib/Controller/LoginController.php`

**Change:**  
Comment the line in `$data += [...]` that sends `claims`:
```php
// 'claims' => json_encode($claims),
```

Then restart Apache.

## 💡 Proposal
- Only send the `claims` parameter if explicitly enabled or populated
- Make `'enable_default_claims' => false` effective
- Or add a setting to completely disable all claim injection

## 🙏 Thanks
If this gets fixed in `user_oidc v7.2.x` or merged into a next release, it will help many Google Workspace users using OIDC with Nextcloud.

Cheers from Spain.

joshtrichards:
Are you sure you don't have some explicit mappings (i.e. for `email` and `name`)? That's the only scenario where those should appear there based on my reading of the code (and that matches the docs):

https://github.com/nextcloud/user_oidc?tab=readme-ov-file#disable-default-claims

julien-nc:
@joshtrichards Thanks for answering this.

@vicentalonso This is indeed very likely that you have explicit mappings for email and name and that's why the claims are added.

If you need to map specific attributes for email and name but you still don't want them to appear in the claim list, this is currently not possible.
I'm pretty sure the OpenID standard tells us to include every requested attribute in the claims. If the google IdP implementation does not work like that, we could introduce a new config value to only include the "user ID" claim...

[...]

PR #1348, opened 2026-02-23
Fix problems with userinfo claims by allowing them to be disabled

carlottostromstedt:
When trying to use the the Application with Google as a provider, the following error is returned:

```
Error 400: invalid_request  
Invalid parameter value for claims: Invalid claims: claims must have a single object value "id_token"
```
To fix this issue i have added the ability to disable userinfo claims by setting `'send_userinfo_claims' => false` in the config.

This adds toggleable compatibility for providers such as Google without breaking current compatibility.

I've tested this fix with my Nextcloud Instance and it fixes the problem at hand. 

--- 
Fixes: https://github.com/nextcloud/user_oidc/issues/1121 Google OIDC login fails in user_oidc v7.2.0 due to unconditionally injected claims

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,6 +186,22 @@ This will use the content of the userinfo endpoint response just like if it had 
 This will only work on login and not when validating a bearer token
 because provisioning when validating a bearer access token is not supported yet.
 
+### Optional userinfo claim
+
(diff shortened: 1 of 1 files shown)


julien-nc:
Rebased on main

PR #1353, opened 2026-02-25
Emit UserCreatedEvent when creating the user in our backend

julien-nc:
Can this have unexpected side effect?

The UserCreatedEvent is listened to in those places:

* https://github.com/nextcloud/server/blob/master/apps/dav/lib/Listener/UserEventsListener.php#L61
* https://github.com/nextcloud/server/blob/master/apps/files_external/lib/Service/MountCacheService.php#L68
* https://github.com/nextcloud/server/blob/master/apps/admin_audit/lib/Listener/UserManagementEventListener.php#L27
* https://github.com/nextcloud/server/blob/master/apps/encryption/lib/Listeners/UserEventsListener.php#L48


julien-nc:
Thank you!

PR #1367, opened 2026-03-10
fix(LoginController): guard dispatch against null user

solracsf:
The event dispatch was called unconditionally, assuming a new user had just been created, causing a fatal TypeError:

`OCP\User\Events\UserCreatedEvent::__construct(): Argument #1 ($user) must be of type OCP\IUser, null given`

Regression from https://github.com/nextcloud/user_oidc/pull/1353

(`$user` will always be null if `'disable_account_creation' => true,`)

### DIFF
--- a/lib/Controller/LoginController.php
+++ b/lib/Controller/LoginController.php
@@ -135,7 +135,7 @@ class LoginController extends BaseOidcController {
 		}
 
 		// Remove protocol and domain name
-		$filtered = preg_replace('/^https?:\/\/[^\/]+/', '', $redirectUrl);
+		$filtered = preg_replace('/^https?:\/\/[^\/]+/', '', $redirectUrl) ?? '';
 
 		// Additional check: ensure the result starts with a single /
 		if (!preg_match('/^\/[^\/]/', $filtered)) {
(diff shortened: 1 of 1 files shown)


lib/Controller/LoginController.php
```diff
@@ -593,10 +607,25 @@ public function code(string $state = '', string $code = '', string $scope = '',
...
+				$disableAccountCreation = isset($oidcSystemConfig['disable_account_creation']) && $oidcSystemConfig['disable_account_creation'];
+				if ($disableAccountCreation) {
+					$this->logger->warning('Account creation is disabled, cannot provision new user', ['userId' => $userId]);
+					// Clean session
+					$this->session->remove(self::STATE);
+					$this->session->remove(self::NONCE);
+					$this->session->remove(self::PROVIDERID);
+					$this->session->remove(self::CODE_VERIFIER);
+					$message = $this->l10n->t('Auto-create users is disabled.');
+					return $this->build403TemplateResponse($message, Http::STATUS_FORBIDDEN, ['reason' => 'account creation disabled'], false);
```
julien-nc: Cleaning the session is nice but returning a 403 here is not exactly needed. It would be done below anyway when checking if there is no user:

``` php
		if ($user === null) {
			$message = $this->l10n->t('Failed to provision the user');
			return $this->build403TemplateResponse($message, Http::STATUS_BAD_REQUEST, ['reason' => 'failed to provision user']);
		}
```

Also why not cleaning the session in all other failing cases?

solracsf:
Reverted all changes beside the real fix & _some code cleanup and error handling_.

julien-nc (approved):
Thank you!

PR #1372, opened 2026-03-12
Fix some deprecations

CarlSchwan:
Keep old compatibility for now, but at least allow us to drop some OC_Util methods at some point


julien-nc:
Thank you!
Psalm is now failing when the new classes are missing in old NC versions.
Would some stubs be ok to fix the psalm errors?

julien-nc:
I hope you don't mind: I have rebased this branch on main and added a fix suggestion.

Since we have the unit/integrations tests that are supposed to fail if we use missing classes/methods, it seems fine to just ignore the `UndefinedClass` and `UndefinedDocblockClass` psalm issues.
We can stop ignoring those when min-nc becomes 34.

CarlSchwan:
Thanks for rebassing, looks fine to me

PR #1375, opened 2026-03-12
fix: handle dot-containing claim names in nested claim resolution

### LINKED ISSUE #1373: Nested claim resolution breaks when claim name is a URL containing dots
## Description

When using `--resolve-nested-and-fallback-claims-mapping=1`, the `getClaimValues()` method in `ProvisioningService.php` splits the claim path by `.` using `explode('.')`. This breaks when the **claim name itself is a URL** containing dots.

This is the case with **EU Login (ECAS)**, the European Commission's SSO provider, which uses URL-based claim names as per OpenID Connect standard practice:

- `https://xxx.xx.xx.x/groups`
- `https://xx.xx.xx.x/claims/extended_attributes`
- `https://xx.xx.xx.xx/claims/domain`

## Steps to reproduce

1. Configure an OIDC provider that returns claims with URL-based names (e.g., EU Login / ECAS)
2. Enable `--resolve-nested-and-fallback-claims-mapping=1`
3. Set `mappingGroups` to `https://xxx.xxx.xx.xxx/extended_attributes`

**Expected:** Access the claim value from the token payload.

**Actual:** `explode('.')` splits the URL into fragments:
```
["https://xxx", "xxxx", "xxx", "xx/claims/extended_attributes"]
```
The parser then tries `$tokenPayload->{"https://xxx"}` which doesn't exist → returns `null`.

## The double problem

There are actually **two separate issues** with dot-based resolution:

### 1. Claim names containing dots (URLs)

The claim name `https://xxxx.xxx.xxx.xxx/claims/extended_attributes` is a flat key in the JWT, but `explode('.')` treats it as a nested path.

### 2. Internal keys containing literal dots

The value of `extended_attributes` is a JSON object whose keys use dot-notation as **literal key names** (not nesting):

```json
{
  "https://xx.xxx.xx.xx/claims/extended_attributes": {
    "user.typeOfActor": "xxxxx",
    "auth.subdomain": "xxxx",
    "auth.permissions": "xxxx"
  }
}
```

`user.typeOfActor` is a flat key, but the parser would interpret it as `user` → `typeOfActor` (two levels of navigation).

## Without resolve-nested-claims

With `--resolve-nested-and-fallback-claims-mapping=0`, the URL-based claim name works as a literal key lookup:
```php
$tokenPayload->{"https://xxx.xx.xx.xxx/claims/extended_attributes"}
```
This returns the object, but there's no way to navigate **into** it to extract a specific attribute like `auth.permissions`.

## Suggested solution

Consider supporting an escape mechanism or bracket notation to distinguish literal dots from path separators. For example:

- **Bracket notation:** `["https://xxx.xxx.xxx.xxx/claims/extended_attributes"]["auth.permissions"]`
- **Escape character:** `https://xxx\.xxx\.xx\.eu/claims/extended_attributes.auth\.permissions`
- **Alternative separator:** Use `/` instead of `.` for path navigation (less breaking)

This is related to #1100 which covers nested claims in general, but this issue specifically addresses the case where **the claim name itself contains dots** (URL-based claim names per OIDC spec) and where **internal keys also contain literal dots**.

## Environment

- Nextcloud: 32.0.6
- user_oidc: latest
- Relevant code: `lib/Service/ProvisioningService.php`, `getClaimValues()` method (lines 67-103)

strobelpierre:
## Summary

When `--resolve-nested-claims=1` is enabled, `getClaimValues()` uses `explode('.')` to split claim paths. This breaks for:

1. **URL-based claim names** (valid per [OIDC Core §5.1.2](https://openid.net/specs/openid-connect-core-1_0.html#AdditionalClaims)): `https://idp.example.com/claims/groups` gets split at hostname dots
2. **Object keys with literal dots**: `{"user.role": "admin"}` gets interpreted as nested navigation

### Approach: greedy longest-prefix matching

Replace `explode('.')` with a recursive resolver (`resolveNestedClaim()`) that:
1. Tries the **full remaining path as a literal key** first
2. Then tries progressively **shorter dot-prefixed segments** (longest first)
3. Recurses into the remainder when a prefix matches a nested object/array

This preserves full backward compatibility — existing dot-separated nested paths resolve identically since the algorithm falls through to the same splits.

### Examples

| Claim path | Token payload | Result |
|---|---|---|
| `https://idp.example.com/claims/groups` | `{"https://...groups": ["admin"]}` | `["admin"]` |
| `https://idp.example.com/attrs.role` | `{"https://...attrs": {"role": "admin"}}` | `"admin"` |
| `https://idp.example.com/attrs.user.role` | `{"https://...attrs": {"user.role": "admin"}}` | `"admin"` |
| `custom.nickname` | `{"custom": {"nickname": "alice"}}` | `"alice"` (backward compat) |
| `a.b` | `{"a.b": "flat", "a": {"b": "nested"}}` | `"flat"` (literal key precedence) |

Fixes #1373
Related: #1100

## Test plan

- [x] Added 13 data-provider test cases covering: flat keys, nested keys, URL claim names, URL + nested navigation, URL + dotted sub-keys, deep nesting, pipe fallbacks, non-existent paths, empty paths, literal dot key precedence, array payloads
- [x] Added regression test for flat mode (nested resolution disabled)
- [ ] CI phpunit pass

### DIFF
--- a/lib/Service/ProvisioningService.php
+++ b/lib/Service/ProvisioningService.php
@@ -87,20 +87,10 @@ class ProvisioningService {
 		$alternatives = explode('|', $claimPath);
 
 		foreach ($alternatives as $altPath) {
-			$parts = explode('.', trim($altPath));
-			$value = $tokenPayload;
-
-			foreach ($parts as $part) {
-				if (is_object($value) && property_exists($value, $part)) {
-					$value = $value->{$part};
-				} elseif (is_array($value) && array_key_exists($part, $value)) {
-					$value = $value[$part];
-				} else {
-					continue 2;
-				}
+			$result = $this->resolveNestedClaim($tokenPayload, trim($altPath));
+			if ($result !== null) {
+				return $result;
 			}
-
-			return $value;
 		}
 
 		return null;
@@ -113,6 +103,50 @@ class ProvisioningService {
 	public function getClaimValue(object|array $tokenPayload, string $claimPath, int $providerId): mixed {
 		$value = $this->getClaimValues($tokenPayload, $claimPath, $providerId);
 		return is_string($value) ? $value : null;
+	}
+
+	/**
+	 * Resolves a claim path against a token payload using greedy longest-prefix matching.
+	 *
+	 * Instead of splitting on every dot (which breaks URL-based claim names like
+	 * "https://idp.example.com/claims/groups" or literal dot keys like "user.role"),
+	 * this method first tries the full path as a literal key, then progressively
+	 * shorter dot-delimited prefixes (longest first), recursing into the remainder.
+	 */
+	private function resolveNestedClaim(object|array $data, string $path): mixed {
+		if ($path === '') {
+			return null;
+		}
+
+		// Try full path as literal key
+		if (is_object($data) && property_exists($data, $path)) {
+			return $data->{$path};
+		} elseif (is_array($data) && array_key_exists($path, $data)) {
+			return $data[$path];
+		}
+
+		// Try progressively shorter dot-prefixes (longest first)
+		$lastDot = strlen($path);
+		while (($lastDot = strrpos($path, '.', -(strlen($path) - $lastDot + 1))) !== false) {
+			$prefix = substr($path, 0, $lastDot);
+			$remainder = substr($path, $lastDot + 1);
+
+			$prefixValue = null;
+			if (is_object($data) && property_exists($data, $prefix)) {
+				$prefixValue = $data->{$prefix};
+			} elseif (is_array($data) && array_key_exists($prefix, $data)) {
+				$prefixValue = $data[$prefix];
+			}
+
+			if ($prefixValue !== null && (is_object($prefixValue) || is_array($prefixValue))) {
+				$result = $this->resolveNestedClaim($prefixValue, $remainder);
+				if ($result !== null) {
+					return $result;
+				}
+			}
+		}
+
+		return null;
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


julien-nc:
@strobelpierre Lgtm. Thank you (and Claude :grin:)

@andreblanke @dragonpil Can you confirm the backward compatibility is preserved in this PR for all your use cases?

strobelpierre:
> @strobelpierre Lgtm. Thank you (and Claude 😁)
> 
> @andreblanke @dragonpil Can you confirm the backward compatibility is preserved in this PR for all your use cases?

Claude and I thank you too, haha 🤝❤️

julien-nc:
@strobelpierre I don't know what's gonna become the standard but here is my advice: Keep the "generated with AI/whatever you use" footer in PR comments and in issues. This is more honest and has less chances to trigger readers that don't like AI and will immediately detect AI generated content from the style and the length.

strobelpierre:
> @strobelpierre I don't know what's gonna become the standard but here is my advice: Keep the "generated with AI/whatever you use" footer in PR comments and in issues. This is more honest and has less chances to trigger readers that don't like AI and will immediately detect AI generated content from the style and the length.

I completely agree with you, but on another project I had a PR that was fine and added a real feature, and I left the mention claude code that I use to formalize the PR format and also to code. The PR was rejected on the grounds that they don't accept code generated by AI, even though all the tests were fine. There are gatekeepers who automatically reject it.

andreblanke:
@julien-nc Looks good to me as well. Thanks for the ping.

julien-nc:
> There are gatekeepers who automatically reject it

I hope this does not turn into a war between AI supporters and detractors.
Honestly your PR is readable, tested, commented, scoped. IMO, good combination of a powerful model and your ability to guide the model wisely.

julien-nc:
@andreblanke Thanks a lot for the super fast feedback!

dragonpil:
Sorry for the late Feedback. Looks good for my usecases.

PR #1376, opened 2026-03-12
fix: set IUserSession user after bearer token validation

hangerrits:
## Summary                                                  
   
  When user_oidc validates a bearer token in Backend::getCurrentUserId(), it returns the correct userId but does not call IUserSession::setUser().    
  This leaves the user session in an inconsistent state where getCurrentUserId() succeeds but DI-injected $userId parameters remain null.          

  ## Problem

  OCS controllers and CalDAV plugins that receive $userId via dependency injection get null instead of the authenticated user's ID when the request
  is authenticated via OIDC bearer token. This causes:

  - **Deck**: TypeError: ...$userId must be of type string, null given
  - **Talk**: Same TypeError pattern
  - **Tasks** (CalDAV): 500 errors from null userId

  These apps work correctly with session-based OIDC login (where setUser() IS called) but fail with bearer token authentication.

  ## Fix

  Call IUserSession::setUser() after successful bearer token validation at all three return points in getCurrentUserId(). IUserSession is resolved via
   Server::get() rather than constructor injection to avoid a circular dependency.

  ## Testing

  1. Configure an OIDC provider with bearer token validation enabled
  2. Make API requests to Deck, Talk, or CalDAV endpoints using a bearer token
  3. Verify 200 responses instead of 500 errors

### DIFF
--- a/lib/User/Backend.php
+++ b/lib/User/Backend.php
@@ -36,6 +36,7 @@ use OCP\ISession;
 use OCP\IURLGenerator;
 use OCP\IUser;
 use OCP\IUserManager;
+use OCP\IUserSession;
 use OCP\Server;
 use OCP\ServerVersion;
 use OCP\User\Backend\ABackend;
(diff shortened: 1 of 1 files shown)


solracsf (approved):

julien-nc (changes requested):
LGTM. Thanks.
If you can rebase on main and fix the cs issue (run `composer run cs:fix`), let's merge that soon.

julien-nc (approved):

artonge:
@hangerrits & @solracsf, looks like this PR is causing the following issue: https://github.com/nextcloud/user_oidc/issues/1452

@pringelmann tried to reproduce your original issue, but couldn't.

I haven't looked too much into it, but could it be that your original issue has been fixed somewhere else, making this PR safely revertable?

solracsf:
@artonge I've approved #1486

artonge:
> @artonge I've approved #1486

Alright,  @pringelmann will take a look at it next week. Maybe @printminion-co can also try out the PR in the meantime and see if it fixes their issue?

PR #1401, opened 2026-04-02
fix: Trigger UserFirstTimeLoggedInEvent

CarlSchwan:
We now depends on NC > 29


julien-nc (commented):
Should we then drop the old one?
``` php
$this->eventDispatcher->dispatch(IUser::class . '::firstLogin', new GenericEvent($user));
```

CarlSchwan:
> Should we then drop the old one?
> 
> ```
> $this->eventDispatcher->dispatch(IUser::class . '::firstLogin', new GenericEvent($user));
> ```

Only once is merged https://github.com/nextcloud/monthly_status_email/pull/233 (and i need to fix the ci)

julien-nc (approved):
Merge whenever you want 😁

PR #1409, opened 2026-04-10
Clarify session values and reauthentication when a token expires

julien-nc:
In some situations, people get the "The received state does not match the expected value." page. This most likely happens because 2 login flows were launched in parallel, one by the user and one by a request that triggers `TokenService::reauthenticate` which redirects to the `user_oidc.login.login` route which starts a login flow that overwrites the session state.

One way to prevent that is to not redirect to `user_oidc.login.login` if we know a login flow might already be in progress.

This PR includes:

* Cleanup session values in `LoginController::code` when we don't need them anymore
* When a token expires, avoid the redirect to the login flow if there is an ongoing login flow (if we see a state in the session)
* Add a config flag to enable a simpler behaviour when a token expires: just logout. This will most likely go away but would be helpful to debug.

#1410 might be a better and more generic fix

### DIFF
--- a/lib/Controller/LoginController.php
+++ b/lib/Controller/LoginController.php
@@ -348,6 +348,7 @@ class LoginController extends BaseOidcController {
 		$this->logger->debug('Code login with core: ' . $code . ' and state: ' . $state);
 
 		if ($error !== '') {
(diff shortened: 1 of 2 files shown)


julien-nc:
Done with #1410

PR #1410, opened 2026-04-10
Prevent parallel login flows after logout

julien-nc:
Global problem: When the user is logged out (by user_oidc or anything else like integration_swp), user_oidc might try to redirect to the start of an Oidc login flow. This can lead to parallel login flows messing with each other (state mismatch) if all requests are redirected.

This can happen:

* after a logout, if `allow_multiple_user_backends` is set to 0 and there is only one provider, any request to /login will redirect to the `user_oidc.login.login` route
* if `store_login_token` is enabled and a token expires, we reach `TokenService::reauthenticate` which logs out and redirects to the `store_login_token`

Solution in this PR: Only redirect to the login flow when the request comes from a 'navigation' context.

This is an alternative to #1409 which is possibly better.

### DIFF
--- a/lib/Service/RequestClassificationService.php
+++ b/lib/Service/RequestClassificationService.php
@@ -1,0 +1,41 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\UserOIDC\Service;
+
+use OCP\IRequest;
+
+class RequestClassificationService {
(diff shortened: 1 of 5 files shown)


CarlSchwan (approved):

lib/Service/RequestClassificationService.php
```diff
@@ -0,0 +1,30 @@
...
+
+		if ($request->getHeader('OCS-apirequest') !== '') {
+			return false;
+		}
+
+		if ($request->getHeader('X-Requested-With') === 'XMLHttpRequest') {
+			return false;
+		}
+
+		return true;
```
CarlSchwan, line 28:
```suggestion
		return $request->getHeader('X-Requested-With') !== 'XMLHttpRequest');
```

PR #1413, opened 2026-04-14
feat: Use new IAlternativeLoginProvider when available

CarlSchwan:
https://github.com/nextcloud/server/pull/59627


lib/AppInfo/Application.php
artonge, line 138: Would add a comment stating that we can remove it once we drop support for NC33 or bellow.

artonge (approved):
Request for a comment, else, looks good

julien-nc (approved):
Nice! No more hacks needed to register multiple ones with IAlternativeLogin! 💙

PR #1414, opened 2026-04-15
perf(ldap): Optimize check if LDAP user is deleted

CarlSchwan:
Check first if the current user is deleted, before loading all the deleted users and checking if their deletion is still up-to-date.

Allow to abort early for non-deleted users.

`isUserMarked` is available since NC 28


julien-nc (approved):
LGTM

PR #1420, opened 2026-04-17
refactor: Harden UserBackend::provisionUser and UserBackend::checkFirstLogin

CarlSchwan:
Add more debug logs when something go wrong and more checks for invalid values.


julien-nc (approved):
💙

PR #1428, opened 2026-04-27
refactor: Trigger UserFirstTimeLoggedIn instead of manually copying skeleton

CarlSchwan:
Need https://github.com/nextcloud/server/pull/59932 first

### DIFF
--- a/lib/Controller/OcsApiController.php
+++ b/lib/Controller/OcsApiController.php
@@ -14,10 +14,12 @@ use OCP\AppFramework\Http;
 use OCP\AppFramework\Http\Attribute\OpenAPI;
 use OCP\AppFramework\Http\DataResponse;
 use OCP\AppFramework\OCSController;
+use OCP\EventDispatcher\IEventDispatcher;
 use OCP\Files\IRootFolder;
 use OCP\Files\NotPermittedException;
 use OCP\IRequest;
 use OCP\IUserManager;
+use OCP\User\Events\UserFirstTimeLoggedInEvent;
 
 class OcsApiController extends OCSController {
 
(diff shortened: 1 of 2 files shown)


lib/Controller/OcsApiController.php
```diff
@@ -26,6 +28,8 @@ public function __construct(
 		private IRootFolder $root,
 		private UserMapper $userMapper,
 		private IUserManager $userManager,
+		private IEventDispatcher $eventDispatcher,
+		private IConfig $config,
```
come-nc, line 32: import is missing.
Also you may want to use ServerVersion?

lib/Controller/OcsApiController.php
```diff
@@ -65,12 +69,19 @@ public function createUser(
...
 
-		$userFolder = $this->root->getUserFolder($user->getUID());
-		try {
-			// copy skeleton
-			\OC_Util::copySkeleton($user->getUID(), $userFolder);
-		} catch (NotPermittedException $ex) {
-			// read only uses
+		// Slighly incorrect, but this event trigger the creation of the file skeleton and
+		// the contact book entry.
+		$this->eventDispatcher->dispatchTyped(new UserFirstTimeLoggedInEvent($user));
```
come-nc, line 74: It is incorrect, we should not trigger this event unless the user logs in.
But with oidc if it’s created it means he’s currently logging in?

julien-nc:
Keep in mind this app supports NC 29 -> 34. I am not against having branches but considering the number of times we have to make fixes/releases for old NC versions, I would recommend not having one branch per server stable.

PR #1431, opened 2026-04-28
fix(login): treat already-logged-out as success in backchannel logout (#1430)

### LINKED ISSUE #1430: RP initiated logout makes backchannel logout throw an HTTP/400 error
### How to use GitHub

---

### Steps to reproduce
1. Logout from the RP (Nextcloud) with backchannel logout enabled on the IdP
2. Agree to disconnect on the IdP interface
3. Backchannel logout will likely throw an error 400

### Expected behaviour
- Do not throw an HTTP/400 error on non existent session [as stated per the RFC](https://openid.net/specs/openid-connect-backchannel-1_0.html#BCActions) 

### Actual behaviour
Backchannel logout will almost every time return an HTTP/400 error when doing a RP-initiated logout.  

### Explaination
The issue is that the error HTTP 400 will then be thrown to the IdP during the backchannel logout due to the [session not being found](https://github.com/nextcloud/user_oidc/blob/4aaac7a76fe8a3765a9020b0bd70be52ccb187d8/lib/Controller/LoginController.php#L933) since the session has been deleted [here](https://github.com/nextcloud/user_oidc/blob/4aaac7a76fe8a3765a9020b0bd70be52ccb187d8/lib/Controller/LoginController.php#L836).
The IdP I am using (LemonLDAP) is registering the error and inform the user a possible error occured. Of course everything is fine  : 

Image

The [RFC does state](https://openid.net/specs/openid-connect-backchannel-1_0.html#BCActions) that a success HTTP/200 should be returned  : 
``` If the identified End-User is already logged out at the RP when the logout request is received, the logout is considered to have succeeded.```

### Proposed fix
- Make the response an HTTP/200 if no session are found ;
- Optionally: have a debug checkbox in the provider option to enable this behavior but have it disabled by default.  The checkbox could be general so it could be use for later use.

I can try to push a PR if you take somebody's first NC related PR :) 

### Server configuration

**Web server:** Nginx

**Database:** PostgreSQL

**PHP version:** 8.4

**Nextcloud version:** 33.0.2

### PR DESCRIPTION
**Pull request by SAY-5**
Closes #1430.

## Summary
Per [OIDC Backchannel Logout 1.0 §2.6](https://openid.net/specs/openid-connect-backchannel-1_0.html#BCActions):

> If the identified End-User is already logged out at the RP when the logout request is received, the logout is considered to have succeeded.

Today `backChannelLogout()` returns **HTTP/400** when the `(sid, iss)` or `(sub, iss)` lookup yields no matching session, which is exactly the already-logged-out case the spec calls out as a **success**. The reporter (running LemonLDAP) sees a confusing error surfaced on the IdP UI even though the desired state — the RP session being gone — is already true. The same shape would happen with Keycloak, Authelia, etc.

## Change
Only the two \"session not found\" paths flip to HTTP/200:

- `findSessionBySid` → `DoesNotExistException` ⇒ return `HTTP/200 OK` + debug log.
- `findSessionsBySubAndIss` returns empty ⇒ return `HTTP/200 OK` + debug log.

The other validation branches in the same method — invalid issuer, invalid signature, malformed token, missing claims, multiple-sessions-found — keep their HTTP/400 responses. Operators that need visibility into already-logged-out backchannel calls can flip Nextcloud's `loglevel` to `debug` and see the `[BackchannelLogout]` line.

Mirrors the proposed fix in the issue and the spec citation.

## Test plan
- No `LoginControllerTest` exists in `tests/unit/Controller/` today, and the existing integration runner under `tests/integration/` does not cover the backchannel flow. The change is two short branches that swap a `getBackchannelLogoutErrorResponse()` for `new JSONResponse([], Http::STATUS_OK)` plus a debug log; both branches are *syntactically* equivalent to the existing `return new JSONResponse([], Http::STATUS_OK)` at the end of the success path.
- Ran `git diff --check` clean.
- Verified the spec citation against the linked RFC section.

## Notes
- This is the spec-compliant behavior; existing IdPs will start receiving cleaner success responses for the already-logged-out case.
- The optional \"debug checkbox\" the reporter floated is left for follow-up — the spec is unambiguous, so a per-provider opt-out adds configuration surface for no clear benefit.

### DIFF
--- a/lib/Controller/LoginController.php
+++ b/lib/Controller/LoginController.php
@@ -930,11 +930,19 @@ class LoginController extends BaseOidcController {
 			try {
 				$oidcSession = $this->sessionMapper->findSessionBySid($sid, $sub, $iss);
 			} catch (DoesNotExistException $e) {
-				return $this->getBackchannelLogoutErrorResponse(
-					$sub === null ? 'invalid SID or ISS' : 'invalid SID, SUB or ISS',
-					$sub === null ? 'No session was found for this (sid,iss)' : 'No session was found for this (sid,sub,iss)',
-					['session_not_found' => $sid]
+				// Per OIDC Backchannel Logout 1.0 §2.6:
+				// "If the identified End-User is already logged out at the
+				// RP when the logout request is received, the logout is
+				// considered to have succeeded." Returning 400 here
+				// causes IdPs (e.g. LemonLDAP, Keycloak) to surface a
+				// confusing error to the user even though the desired
+				// state — the RP session being gone — is already true.
+				// See https://openid.net/specs/openid-connect-backchannel-1_0.html#BCActions
+				$this->logger->debug(
+					'[BackchannelLogout] no RP session for (sid,iss) — treating as already-logged-out per spec',
+					['sid' => $sid, 'sub_present' => $sub !== null]
 				);
+				return new JSONResponse([], Http::STATUS_OK);
 			} catch (MultipleObjectsReturnedException $e) {
 				return $this->getBackchannelLogoutErrorResponse(
 					$sub === null ? 'invalid SID or ISS' : 'invalid SID, SUB or ISS',
@@ -957,11 +965,14 @@ class LoginController extends BaseOidcController {
 			}
 
 			if (empty($oidcSessionsToKill)) {
-				return $this->getBackchannelLogoutErrorResponse(
-					'nothing found with sub+iss',
-					'No session found with sub+iss',
-					['sub_iss_no_session_found' => true]
+				// Per OIDC Backchannel Logout 1.0 §2.6 (already-logged-
+				// out is a success). Same rationale as the (sid,iss)
+				// branch above.
+				$this->logger->debug(
+					'[BackchannelLogout] no RP sessions for (sub,iss) — treating as already-logged-out per spec',
+					['sub' => $sub]
 				);
+				return new JSONResponse([], Http::STATUS_OK);
 			}
 		}
 
--- a/lib/Controller/LoginController.php
+++ b/lib/Controller/LoginController.php
@@ -930,16 +930,10 @@ class LoginController extends BaseOidcController {
 			try {
 				$oidcSession = $this->sessionMapper->findSessionBySid($sid, $sub, $iss);
 			} catch (DoesNotExistException $e) {
-				// Per OIDC Backchannel Logout 1.0 §2.6:
-				// "If the identified End-User is already logged out at the
-				// RP when the logout request is received, the logout is
-				// considered to have succeeded." Returning 400 here
-				// causes IdPs (e.g. LemonLDAP, Keycloak) to surface a
-				// confusing error to the user even though the desired
-				// state — the RP session being gone — is already true.
-				// See https://openid.net/specs/openid-connect-backchannel-1_0.html#BCActions
+				// Already-logged-out is a success per OIDC Backchannel Logout 1.0 §2.6.
+				// https://openid.net/specs/openid-connect-backchannel-1_0.html#BCActions
 				$this->logger->debug(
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by Spitfireap** [comment]
Thanks ! 
tested on v33.0.2 without any issue. It's working as intended

File: lib/Controller/LoginController.php [hunk 1]
```diff
@@ -957,11 +965,14 @@ public function backChannelLogout(string $providerIdentifier, string $logout_tok
...
 			if (empty($oidcSessionsToKill)) {
-				return $this->getBackchannelLogoutErrorResponse(
-					'nothing found with sub+iss',
-					'No session found with sub+iss',
-					['sub_iss_no_session_found' => true]
+				// Per OIDC Backchannel Logout 1.0 §2.6 (already-logged-
+				// out is a success). Same rationale as the (sid,iss)
+				// branch above.
+				$this->logger->debug(
+					'[BackchannelLogout] no RP sessions for (sub,iss) — treating as already-logged-out per spec',
```
```suggestion
					'[BackchannelLogout] no RP sessions for (sub,iss) — treating as already-logged-out',
```

lib/Controller/LoginController.php
```diff
@@ -930,11 +930,19 @@ public function backChannelLogout(string $providerIdentifier, string $logout_tok
...
+				// Per OIDC Backchannel Logout 1.0 §2.6:
+				// "If the identified End-User is already logged out at the
+				// RP when the logout request is received, the logout is
+				// considered to have succeeded." Returning 400 here
+				// causes IdPs (e.g. LemonLDAP, Keycloak) to surface a
+				// confusing error to the user even though the desired
+				// state — the RP session being gone — is already true.
+				// See https://openid.net/specs/openid-connect-backchannel-1_0.html#BCActions
+				$this->logger->debug(
+					'[BackchannelLogout] no RP session for (sid,iss) — treating as already-logged-out per spec',
```
```suggestion
					'[BackchannelLogout] no RP session for (sid,iss) — treating as already-logged-out',
```

lib/Controller/LoginController.php
```diff
@@ -930,11 +930,19 @@ public function backChannelLogout(string $providerIdentifier, string $logout_tok
...
-					$sub === null ? 'No session was found for this (sid,iss)' : 'No session was found for this (sid,sub,iss)',
-					['session_not_found' => $sid]
+				// Per OIDC Backchannel Logout 1.0 §2.6:
+				// "If the identified End-User is already logged out at the
+				// RP when the logout request is received, the logout is
+				// considered to have succeeded." Returning 400 here
+				// causes IdPs (e.g. LemonLDAP, Keycloak) to surface a
+				// confusing error to the user even though the desired
+				// state — the RP session being gone — is already true.
+				// See https://openid.net/specs/openid-connect-backchannel-1_0.html#BCActions
```
julien-nc: Can you make this a little bit shorter? The link + a very brief explanation should be enough.

julien-nc (changes requested):
Thanks a lot. Small adjustments and let's get this in!

SAY-5:
Thanks for the review @julien-nc — applied all three suggestions in [aa87b8f](https://github.com/nextcloud/user_oidc/pull/1431/commits/aa87b8f): comment block shortened to a single line + spec link, and the redundant `' per spec'` suffix dropped from both debug log messages.

julien-nc (approved):

julien-nc:
Thanks for the adjustments. Could you sign your commits like indicated in the DCO check output?

SAY-5:
@julien-nc the DCO check is currently passing — both commits on this branch carry `Signed-off-by: SAY-5 <SAY-5@users.noreply.github.com>`. Latest run: ✅ `DCO` (https://github.com/nextcloud/user_oidc/actions). If the maintainer-side merge UI is showing the GitHub commit-verification badge as unverified, that's a separate signing concern — happy to re-push with SSH-verified signatures if that's the blocker. Let me know which one you saw failing.

julien-nc:
Thank you!

PR #1432, opened 2026-04-28
Align with backchannel logout OIDC spec

Spitfireap:
# Current behaviour
The response of BackChannel logout seems to be missing the Cache-Control header ([spec](https://openid.net/specs/openid-connect-backchannel-1_0.html#BCResponse)).

The BC logout validation is also missing exp token validation (not expired) and iss validation (equals to issuer in discovery endpoint) per [Backchannel logout token validation spec](https://openid.net/specs/openid-connect-backchannel-1_0.html#Validation) and [ID Token validation spec](https://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation)

# Changes
- Removed an unused variable from `getBackchannelLogoutErrorResponse` (`$throttleMetadata`)
- Increased the severity of the logging when a Backchannel logout fails 
- Added the Cache-Control header in the backchannel logout response
- Added a verification for the exp and iss claim
- Check that all required tokens by the spec are there

### DIFF
--- a/lib/Controller/LoginController.php
+++ b/lib/Controller/LoginController.php
@@ -889,7 +889,8 @@ class LoginController extends BaseOidcController {
 
 		// REQUIRED claims check step
 		// https://openid.net/specs/openid-connect-backchannel-1_0.html#LogoutToken
-		$requiredClaims = ['iss', 'aud', 'iat', 'exp', 'jti', 'events'];
+		// Note : exp claim is deliberately not checked. See #1432
+		$requiredClaims = ['iss', 'aud', 'iat', 'jti', 'events'];
 		$missingClaims = [];
 		$logoutTokenArray = (array)$logoutTokenPayload;
 		foreach ($requiredClaims as $claim) {
@@ -959,7 +960,7 @@ class LoginController extends BaseOidcController {
 			);
 		}
 
-		if ($logoutTokenPayload->exp < $this->timeFactory->getTime()) {
+		if (isset($logoutTokenPayload->exp) && $logoutTokenPayload->exp < $this->timeFactory->getTime()) {
 			return  $this->getBackchannelLogoutErrorResponse(
 				'invalid exp',
 				'The logout token is expired',
--- a/lib/Controller/LoginController.php
+++ b/lib/Controller/LoginController.php
@@ -877,6 +877,27 @@ class LoginController extends BaseOidcController {
 
 		$this->logger->debug('Parsed the logout JWT payload: ' . json_encode($logoutTokenPayload, JSON_THROW_ON_ERROR));
 
+		// REQUIRED claims check step
+		// https://openid.net/specs/openid-connect-backchannel-1_0.html#LogoutToken
+		$requiredClaims = ['iss', 'aud', 'iat', 'exp', 'jti', 'events'];
+		$missingClaims = [];
+		$logoutTokenArray = (array)$logoutTokenPayload;
+		foreach ($requiredClaims as $claim) {
+			if (!array_key_exists($claim, $logoutTokenArray)) {
+				$missingClaims[] = $claim;
+			}
+		}
+		if (!empty($missingClaims)) {
+			return $this->getBackchannelLogoutErrorResponse(
+				'missing one or more claims',
+				'missing the following claim(s) : ' . implode(', ', $missingClaims),
+				['extra_context' => 'Probably is an IdP side issue']
+			);
+		}
+
+		// Logout token validation step
+		// https://openid.net/specs/openid-connect-backchannel-1_0.html#Validation
+
 		// check the audience
 		$aud = $logoutTokenPayload->aud;
 		$clientId = $provider->getClientId();
@@ -894,7 +915,7 @@ class LoginController extends BaseOidcController {
 		}
 
 		// check the event attr
-		if (!isset($logoutTokenPayload->events->{'http://schemas.openid.net/event/backchannel-logout'})) {
+		if (!$logoutTokenPayload->events->{'http://schemas.openid.net/event/backchannel-logout'}) {
 			return $this->getBackchannelLogoutErrorResponse(
 				'invalid event',
 				'The backchannel-logout event was not found in the logout token',
@@ -914,18 +935,7 @@ class LoginController extends BaseOidcController {
 			);
 		}
 
-		if (!isset($logoutTokenPayload->iss)) {
-			return $this->getBackchannelLogoutErrorResponse(
-				'invalid iss',
-				'The logout token should contain an iss attribute',
-				[
-					'extra_context' => 'Probably is an IdP side issue',
-					'iss' => $logoutTokenPayload->iss,
-				],
-			);
-		}
 		$iss = $logoutTokenPayload->iss;
-
 		if (!isset($logoutTokenPayload->sid) && !isset($logoutTokenPayload->sub)) {
 			return $this->getBackchannelLogoutErrorResponse(
 				'invalid sid+sub',
--- a/lib/Controller/LoginController.php
+++ b/lib/Controller/LoginController.php
@@ -846,6 +846,9 @@ class LoginController extends BaseOidcController {
 	 * Endpoint called by the IdP (OP) when end_session_endpoint is called by another client
 	 * The logout token contains the sid for which we know the sessionId
 	 * which leads to the auth token that we can invalidate
+	 * Note : in a RP-initiated logout scenario
+	 * the invalidation step should not be required since it would have been cleared
+	 * in singleLogoutService()
 	 * Implemented according to https://openid.net/specs/openid-connect-backchannel-1_0.html
 	 *
 	 * @param string $providerIdentifier
@@ -863,7 +866,7 @@ class LoginController extends BaseOidcController {
 			return $this->getBackchannelLogoutErrorResponse(
 				'provider not found',
 				'The provider was not found in Nextcloud',
-				['provider_not_found' => $providerIdentifier]
+				['extra_context' => 'Got provider identifier: ' . $providerIdentifier],
 			);
 		}
 
@@ -882,7 +885,11 @@ class LoginController extends BaseOidcController {
 			return $this->getBackchannelLogoutErrorResponse(
 				'invalid audience',
 				'The audience of the logout token does not match the provider',
-				['invalid_audience' => $logoutTokenPayload->aud]
+				[
+					'extra_context' => 'Probably is an IdP side issue',
+					'aud' => $aud,
+					'client_id' => $clientId,
+				]
 			);
 		}
 
@@ -891,7 +898,10 @@ class LoginController extends BaseOidcController {
 			return $this->getBackchannelLogoutErrorResponse(
 				'invalid event',
 				'The backchannel-logout event was not found in the logout token',
-				['invalid_event' => true]
+				[
+					'extra_context' => 'Probably is an IdP side issue',
+					'events' => $logoutTokenPayload->events,
+				]
 			);
 		}
 
@@ -900,7 +910,7 @@ class LoginController extends BaseOidcController {
 			return $this->getBackchannelLogoutErrorResponse(
 				'invalid nonce',
 				'The logout token should not contain a nonce attribute',
-				['nonce_should_not_be_set' => true]
+				['extra_context' => 'Probably is an IdP side issue'],
 			);
 		}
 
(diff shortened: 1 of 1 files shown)


Spitfireap:
@julien-nc not entirely sure about the change for the error logging.
I think the severity should be increased since it might be a security issue. The thing is : could these be triggered by a random idp and in that case we should perhaps be cautious about using error or it requires the idp to be verified so error would be well suited ?

lib/Controller/LoginController.php
```diff
@@ -929,39 +929,17 @@ public function backChannelLogout(string $providerIdentifier, string $logout_tok
...
-				);
-			}
-
-			if (empty($oidcSessionsToKill)) {
-				return $this->getBackchannelLogoutErrorResponse(
-					'nothing found with sub+iss',
-					'No session found with sub+iss',
-					['sub_iss_no_session_found' => true]
-				);
+			} catch (DoesNotExistException $e) {
```
julien-nc: `DoesNotExistException` is not thrown by `findEntities` inside `findSessionsBySubAndIss`. Let's keep those 2 blocks there please (the catch and the "if empty") and make them only log something.

lib/Controller/LoginController.php
```diff
@@ -929,39 +929,17 @@ public function backChannelLogout(string $providerIdentifier, string $logout_tok
...
 			try {
 				$oidcSession = $this->sessionMapper->findSessionBySid($sid, $sub, $iss);
+				$oidcSessionsToKill[] = $oidcSession;
 			} catch (DoesNotExistException $e) {
-				return $this->getBackchannelLogoutErrorResponse(
-					$sub === null ? 'invalid SID or ISS' : 'invalid SID, SUB or ISS',
-					$sub === null ? 'No session was found for this (sid,iss)' : 'No session was found for this (sid,sub,iss)',
-					['session_not_found' => $sid]
-				);
-			} catch (MultipleObjectsReturnedException $e) {
```
julien-nc, line 940: Please keep this catch. It is safer to catch `MultipleObjectsReturnedException` even if it is not supposed to happen.
`findEntity` can throw it so we just catch it.

julien-nc (changes requested):
Nice, thanks. Let's get #1431 in first, then you can rebase and adjust this one.

lib/Controller/LoginController.php
Spitfireap: Got it, fixed

Spitfireap:
Added some more alignment with the spec (exp + 1 iss validation step).

Was wondering if we should verify that `jti` and `iat` claims are there as well ([since they are required](https://openid.net/specs/openid-connect-backchannel-1_0.html#LogoutToken)) or am I overdoing it ?

lib/Controller/LoginController.php
```diff
@@ -900,24 +914,39 @@ public function backChannelLogout(string $providerIdentifier, string $logout_tok
...
 		}
 
 		if (!isset($logoutTokenPayload->iss)) {
 			return $this->getBackchannelLogoutErrorResponse(
 				'invalid iss',
 				'The logout token should contain an iss attribute',
-				['iss_should_be_set' => true]
+				true
+			);
+		} elseif ($iss !== $discovery['issuer']) {
```
julien-nc: `$iss` is not defined at this point.

lib/Controller/LoginController.php
```diff
@@ -1001,16 +1019,30 @@ public function backChannelLogout(string $providerIdentifier, string $logout_tok
...
 		string $description,
-		array $throttleMetadata = [],
+		bool $isLikelyIdpSide,
 	): JSONResponse {
-		$this->logger->debug('Backchannel logout error. ' . $error . ' ; ' . $description);
-		return new JSONResponse(
+		// Inform admins that the backchannel logout didn't work because of a misconfiguration
+		if ($isLikelyIdpSide) {
+			$this->logger->error('Backchannel logout error. ' . $error . ' ; ' . $description .
+			'. This is likely an IdP issue.');
```
```suggestion
			$this->logger->error('Backchannel logout error. ' . $error . ' ; ' . $description .
			'. This is likely an IdP issue.', ['metadata' => $metadata]);
```

You could keep the metadata param (which can be renamed) and include it in the logs so the information is more precise than the `$isLikelyIdpSide` boolean. Wdyt?

lib/Controller/LoginController.php
```diff
@@ -928,40 +957,24 @@ public function backChannelLogout(string $providerIdentifier, string $logout_tok
...
-				);
+				$this->logger->debug('[BackchannelLogout] OIDC session not found with sid+sub+iss (expected for a RP-initiated logout)');
 			} catch (MultipleObjectsReturnedException $e) {
-				return $this->getBackchannelLogoutErrorResponse(
-					$sub === null ? 'invalid SID or ISS' : 'invalid SID, SUB or ISS',
-					$sub === null ? 'Multiple sessions were found with this (sid,iss)' : 'Multiple sessions were found with this (sid,sub,iss)',
-					['multiple_sessions_found' => $sid]
-				);
+				$this->logger->warning('[BackchannelLogout] Multiple OIDC sessions retrieved (sid+sub+iss). ' .
+				'This should not happen. Please check that you have created your DB indexes')
```
julien-nc: Missing `;`

julien-nc (changes requested):
Hey. #1431 is merged. Sorry there seems to be a small conflict with your branch. You can rebase on main or just merge main in your branch to resolve the conflict.

> Was wondering if we should verify that jti and iat claims are there as well...

Yeah sure, let's keep it simple: It's enough to check if those token attributes are defined.

Spitfireap:
> Hey. #1431 is merged. Sorry there seems to be a small conflict with your branch. You can rebase on main or just merge main in your branch to resolve the conflict.
> 
> > Was wondering if we should verify that jti and iat claims are there as well...
> 
> Yeah sure, let's keep it simple: It's enough to check if those token attributes are defined.

Working on May the 1st are you :p ? 
I'll push the rebase and some rework, then test it a bit on my instance. I'll ping you when it's ready.

julien-nc:
> Working on May the 1st are you :p ?

Checking your github notifications on May 1st? :grin: 

Yeah for a weird reason I'm working today. I can tell you on a private channel if you're curious.

> I'll push the rebase and some rework, then test it a bit on my instance. I'll ping you when it's ready.

Nice! Thank you for the enthusiasm. No rush though, feel free to take the public holiday :wink:

Spitfireap:
@julien-nc should be ready.
Funny thing : LemonLDAP implementation of OP [is actually not aligned with the spec](https://gitlab.ow2.org/lemonldap-ng/lemonldap-ng/-/work_items/3609) (exp token missing). 
So although my initial issue is fixed, I still get the error message after this PR. I don't think we should let it pass anyway though. Wdyt ?

Spitfireap:
Note : this means that LemonLDAP's users won't be logged out anymore in case of OP initiated logout...
This might be a security issue so idk if we should add a way of bypassing it or add a warning of some sort in the README for instance

lib/Controller/LoginController.php
```diff
@@ -998,22 +1045,29 @@ public function backChannelLogout(string $providerIdentifier, string $logout_tok
...
-	 * @param array $throttleMetadata
+	 * @param array $metadata
 	 * @return JSONResponse
 	 */
 	private function getBackchannelLogoutErrorResponse(
 		string $error,
 		string $description,
-		array $throttleMetadata = [],
+		array $metadata = [],
+		string $severity = 'warning',
```
```suggestion
		string $severity = \Psr\Log\LogLevel::WARNING,
```

lib/Controller/LoginController.php
```diff
@@ -863,7 +866,17 @@ public function backChannelLogout(string $providerIdentifier, string $logout_tok
...
+			);
+		}
+
+		try {
+			$discovery = $this->discoveryService->obtainDiscovery($provider);
+		} catch (\Exception $e) {
+			return $this->getBackchannelLogoutErrorResponse(
+				'could not reach provider endpoint',
+				'URL: ' . $provider->getDiscoveryEndpoint() . 'was not reachable',
+				severity: 'error',
```
```suggestion
				severity: \Psr\Log\LogLevel::ERROR,
```

lib/Controller/LoginController.php
```diff
@@ -882,16 +916,19 @@ public function backChannelLogout(string $providerIdentifier, string $logout_tok
 			return $this->getBackchannelLogoutErrorResponse(
 				'invalid audience',
 				'The audience of the logout token does not match the provider',
-				['invalid_audience' => $logoutTokenPayload->aud]
+				['extra_context' => 'Probably is an IdP side issue', 'aud' => $aud]
```
```suggestion
				['extra_context' => 'Probably is an IdP side issue', 'aud' => $aud, 'client_id' => $clientId]
```

lib/Controller/LoginController.php
```diff
@@ -900,24 +937,40 @@ public function backChannelLogout(string $providerIdentifier, string $logout_tok
...
 				'invalid nonce',
 				'The logout token should not contain a nonce attribute',
-				['nonce_should_not_be_set' => true]
+				['extra_context' => 'Probably is an IdP side issue']
 			);
 		}
 
-		if (!isset($logoutTokenPayload->iss)) {
+		$iss = $logoutTokenPayload->iss;
+		if ($iss !== $discovery['issuer']) {
```
```suggestion
		$iss = $logoutTokenPayload->iss;
		$discoveryIssuer = $discovery['issuer'] ?? '';
		if ($iss !== $discoveryIssuer) {
```

lib/Controller/LoginController.php
```diff
@@ -900,24 +937,40 @@ public function backChannelLogout(string $providerIdentifier, string $logout_tok
...
 			return $this->getBackchannelLogoutErrorResponse(
 				'invalid iss',
-				'The logout token should contain an iss attribute',
-				['iss_should_be_set' => true]
+				'The iss of the logout token does not match the issuer',
+				[
+					'extra_context' => 'Probably is an IdP side issue',
+					'iss' => $iss,
+					'issuer' => $discovery['issuer']
```
```suggestion
					'issuer' => $discoveryIssuer,
```

lib/Controller/LoginController.php
```diff
@@ -900,24 +937,40 @@ public function backChannelLogout(string $providerIdentifier, string $logout_tok
...
+		}
+
+		if ($logoutTokenPayload->exp < $this->timeFactory->getTime()) {
+			return  $this->getBackchannelLogoutErrorResponse(
+				'invalid exp',
+				'The logout token is expired',
+				[
+					'extra_context' => 'Probably is an IdP side issue',
+					'exp' => $logoutTokenPayload->exp,
+					'current_time' => $this->timeFactory->getTime()
```
```suggestion
					'current_time' => $this->timeFactory->getTime(),
```

lib/Controller/LoginController.php
```diff
@@ -929,42 +982,31 @@ public function backChannelLogout(string $providerIdentifier, string $logout_tok
...
-				);
-				return new JSONResponse([], Http::STATUS_OK);
+				$this->logger->debug('[BackchannelLogout] OIDC session not found with sid+sub+iss (expected for a RP-initiated logout)');
 			} catch (MultipleObjectsReturnedException $e) {
-				return $this->getBackchannelLogoutErrorResponse(
-					$sub === null ? 'invalid SID or ISS' : 'invalid SID, SUB or ISS',
-					$sub === null ? 'Multiple sessions were found with this (sid,iss)' : 'Multiple sessions were found with this (sid,sub,iss)',
-					['multiple_sessions_found' => $sid]
+				$this->logger->warning('[BackchannelLogout] Multiple OIDC sessions retrieved (sid+sub+iss). '
+				. 'This should not happen. Please check that you have created your DB indexes'
```
```suggestion
				. 'This should not happen.'
```
Missing indexes don't change anything to this. It cannot make the request return multiple results.

julien-nc (changes requested):

julien-nc:
> I don't think we should let it pass anyway though. Wdyt ?

I think it's fine to let this pass if the token expiration is not defined. So:
* If the exp is defined, we validate it
* If it's not defined, we just ignore

Sounds good to you?

Spitfireap:
> > I don't think we should let it pass anyway though. Wdyt ?
> 
> I think it's fine to let this pass if the token expiration is not defined. So:
> 
>     * If the exp is defined, we validate it
> 
>     * If it's not defined, we just ignore
> 
> 
> Sounds good to you?

If you say so. Although it is a clear deviation from the spec I think since it is REQUIRED, and we ought to verify it as we would do in the code function...

For instance, [had LemonLDAP not forget to make the exp claim required in its RP implementation, the OP would get the 400 error](https://gitlab.ow2.org/lemonldap-ng/lemonldap-ng/-/commit/8ba8d37eea90ab41572ad5b0e828b9e92c9f1336?merge_request_iid=959) pretty much as current fix.

So do I make it optional ?

julien-nc (approved):
Let's get this in while it implements the RFC. We can discuss the missing exp later.

julien-nc:
Thanks a lot for the adjustments.

PR #1441, opened 2026-05-18
Fix: Make exp claim required for back-channel logout

### LINKED ISSUE #1436: Backchannel logout missing exp claim check
### How to use GitHub

---

### Steps to reproduce
1. Send a logout token without the exp claim through an IdP through Backchannel logout

### Expected behaviour
The IdP should recieve an HTTP/400 error since the exp claim [is REQUIRED in the logout token](https://openid.net/specs/openid-connect-backchannel-1_0.html#LogoutToken). It should be [validated](https://openid.net/specs/openid-connect-backchannel-1_0.html#Validation) as usual.

### Actual behaviour
A success is yield if the exp claim is missing from the logout token

### Note
Discussed in #1432. This issue is for visibility and tracking. The exp claim is not made required yet since it would prevent LemonLDAP from logging out user through backchannel logout ([upstream issue](https://gitlab.ow2.org/lemonldap-ng/lemonldap-ng/-/work_items/3609)).
The commit 7fcb03d should be reverted when the issue is fixed upstream.
I'll update this issue when it is done so.

Spitfireap:
fixes #1436 

This reverts commit 7fcb03d5cb20a4cb7ed8e076a1fd5a4e19d8232d.

The exp claim in the logout token during back-channel logout was not checked due to LemonLDAP (third party IdP) not including it. LemonLDAP's users would have been unable to logout with back-channel.

Now that the fix is merged, the commit can be reverted.

### DIFF
--- a/lib/Controller/LoginController.php
+++ b/lib/Controller/LoginController.php
@@ -959,7 +959,7 @@ class LoginController extends BaseOidcController {
 			);
 		}
 
-		if ($logoutTokenPayload->exp < $this->timeFactory->getTime()) {
+		if (($logoutTokenPayload->exp ?? 0) < $this->timeFactory->getTime()) {
 			return  $this->getBackchannelLogoutErrorResponse(
 				'invalid exp',
 				'The logout token is expired',
--- a/lib/Controller/LoginController.php
+++ b/lib/Controller/LoginController.php
@@ -889,8 +889,7 @@ class LoginController extends BaseOidcController {
 
 		// REQUIRED claims check step
 		// https://openid.net/specs/openid-connect-backchannel-1_0.html#LogoutToken
-		// Note : exp claim is deliberately not checked. See #1432
-		$requiredClaims = ['iss', 'aud', 'iat', 'jti', 'events'];
+		$requiredClaims = ['iss', 'aud', 'iat', 'exp', 'jti', 'events'];
 		$missingClaims = [];
 		$logoutTokenArray = (array)$logoutTokenPayload;
 		foreach ($requiredClaims as $claim) {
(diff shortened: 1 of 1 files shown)


lib/Controller/LoginController.php
```diff
@@ -960,7 +959,7 @@ public function backChannelLogout(string $providerIdentifier, string $logout_tok
 			);
 		}
 
-		if (isset($logoutTokenPayload->exp) && $logoutTokenPayload->exp < $this->timeFactory->getTime()) {
+		if ($logoutTokenPayload->exp < $this->timeFactory->getTime()) {
```
```suggestion
		if (($logoutTokenPayload->exp ?? 0) < $this->timeFactory->getTime()) {
```
Let's handle the case where it is not defined to avoid unpredictable behaviours.

julien-nc (changes requested):
Thank you!

lib/Controller/LoginController.php
Spitfireap: > Let's handle the case where it is not defined to avoid unpredictable behaviours.

Ok, wouldn't it be better to catch null values [here](https://github.com/nextcloud/user_oidc/blob/7be6bb6fb80779c4d8ba9e5fb3966a32cc518f3f/lib/Controller/LoginController.php#L896) ? 
Or is it a failsafe for clarity and robustness against later development ?

julien-nc: As you wish. My goal was to keep the code short.

Spitfireap: Must be me, but I fail to see any scenario where this might change the result of the comparison. I'd be glad if you had time to explain it :).
Anyway, done

julien-nc (approved):

lib/Controller/LoginController.php
julien-nc: If the exp attribute is not defined then we use 0 so we know the comparison with getTime will fail so we actually require this attribute. Am I missing something?

Your initial version might have had the exact same behaviour but is less explicit.

Spitfireap: > If the exp attribute is not defined then we use 0 so we know the comparison with getTime will fail so we actually require this attribute. Am I missing something?
> 
> Your initial version might have had the exact same behaviour but is less explicit.

For me it cannot be undefined because it is checked at line 892 ($requiredClaims) and it will return (l.901) before if missing (undefined).
Worse case scenario is that it is null or empty string but the int casting will make it 0.

PR #1443, opened 2026-05-22
Fixed Disable Other Login Methods command

JATippit:
Fixed Disable Other Login Methods command.  Missed a /


julien-nc (approved):

julien-nc:
Thank you!

PR #1444, opened 2026-05-22
Allow setting `enrich_login_id_token_with_userinfo` per-provider in the UI

s3n-w6i:
**What**

This change introduces a checkbox into the provider settings that allows enabling the existing `enrich_login_id_token_with_userinfo` option on a per-provider basis.
To be backwards compatible, if the global `enrich_login_id_token_with_userinfo` is set to true, the userinfo endpoint is still fetched for all providers, regardless of their per-provider setting.

**Why**

- It depends on the provider whether the userinfo endpoint needs to be fetched, so a per-provider setting better represents that.
- For people with a hosted Nextcloud offering like myself, there may not be an option to set the global configuration option.

**Open questions**

- I have tested it successfully using a Keycloak instance. Are integration tests required?
- Regarding migration: This is a new option `enrichLoginIdTokenWithUserinfo` on the provider's settings object. This does not require any database schema changes, right?


s3n-w6i:
Closing to rework this on another branch with proper sign-off commits

PR #1445, opened 2026-05-22
Add per-provider enrich with userinfo endpoint setting

s3n-w6i:
**What**

This change introduces a checkbox into the provider settings that allows enabling the existing `enrich_login_id_token_with_userinfo` option on a per-provider basis.
To be backwards compatible, if the global `enrich_login_id_token_with_userinfo` is set to true, the userinfo endpoint is still fetched for all providers, regardless of their per-provider setting.

**Why**

- It depends on the provider whether the userinfo endpoint needs to be fetched, so a per-provider setting better represents that.
- For people with a hosted Nextcloud offering like myself, there may not be an option to set the global configuration option.

**Open questions**

- I have tested it successfully using a Keycloak instance. Are integration tests required?
- Regarding migration: This is a new option `enrichLoginIdTokenWithUserinfo` on the provider's settings object. This does not require any database schema changes, right?


l10n/en_GB.json
```diff
@@ -112,6 +112,8 @@
     "Should the ID token be included as the id_token_hint GET parameter in the OpenID logout URL? Users are redirected to this URL after logging out of Nextcloud. Enabling this setting exposes the OIDC ID token to the user agent, which may not be necessary depending on the OIDC provider." : "Should the ID token be included as the id_token_hint GET parameter in the OpenID logout URL? Users are redirected to this URL after logging out of Nextcloud. Enabling this setting exposes the OIDC ID token to the user agent, which may not be necessary depending on the OIDC provider.",
```
julien-nc, line 112: You don't need to update the translations. The source file is generated by the translation bot and the translations are produced in transifex and synced by the bot. You can revert the changes in this file.

julien-nc (changes requested):
Nice PR, thanks.
* Backward compatibility 👍 
* No migration needed 👍 because the value is stored in app config as a new row with the key `provider-N-enrichLoginIdTokenWithUserinfo` for each provider

l10n/en_GB.json
s3n-w6i: Ok, I have reverted them 👍

julien-nc (approved):

PR #1478, opened 2026-07-22
fix: Check group allow list based on the display name

### LINKED ISSUE #1472: Users are not automatically removed from Keycloak groups
My Nextcloud version is: 34.0.1

My user_oidc version is: 8.10.1

### Steps to reproduce
1. Configure Nextcloud with user_oidc to provision users and groups from Keycloak
2. Enable unique user IDs
3. Use a regular expressions such as /^wolke/ to filter groups

### Expected behaviour

When users are added to or removed from groups in Keycloak a users should be added or removed from a group when they log in, depending on whether they are a member of the group.

### Actual behaviour
Users are added to groups, but never removed from groups.

### Debugging results

The removal code, that should remove users from groups only considers the group ID, not the group name, so the regular expression is useless, because it works on a random ID.

This patch fixed the issue for me:

```patch
--- a/lib/Service/ProvisioningService.php
+++ b/lib/Service/ProvisioningService.php
@@ -758,7 +758,8 @@
 		$userGroups = $this->groupManager->getUserGroups($user);
 		foreach ($userGroups as $group) {
 			if (!in_array($group->getGID(), array_column($syncGroups, 'gid'))) {
-				if ($groupsWhitelistRegex && !preg_match($groupsWhitelistRegex, $group->getGID())) {
+				$groupNameToMatch = $group->getDisplayName() ?? $group->getGID();
+				if ($groupsWhitelistRegex && !preg_match($groupsWhitelistRegex, $groupNameToMatch)) {
 					continue;
 				}
 				$group->removeUser($user);
``` 

The patch seems not to be compatible with the current master branch, but the issue should be fixed in upstream.

Thanks!

CarlSchwan:
Since a few versions ago, we use ICreateNamedGroupBackend to create group on the default database backend. This will use the display name as GID unless the display name is too long and in that case will use a random GID.

This random GID is impossible to match with the regex.

Fix #1472 

Thanks to @OScaleSven for the initial research :)

## 🤖 AI (if applicable)

- [ ] The content of this PR was partly or fully generated using AI

### DIFF
--- a/lib/Service/ProvisioningService.php
+++ b/lib/Service/ProvisioningService.php
@@ -762,13 +762,22 @@ class ProvisioningService {
 			return null;
 		}
 
-		$userGroups = $this->groupManager->getUserGroupIds($user);
-		foreach ($userGroups as $groupGID) {
-			if (!in_array($groupGID, array_column($syncGroups, 'gid'))) {
-				if ($groupsWhitelistRegex && !preg_match($groupsWhitelistRegex, $groupGID)) {
-					continue;
+		if ($groupsWhitelistRegex) {
+			$userGroups = $this->groupManager->getUserGroups($user);
+			foreach ($userGroups as $group) {
+				if (!in_array($group->getGID(), array_column($syncGroups, 'gid'))) {
+					if (!preg_match($groupsWhitelistRegex, $group->getDisplayName())) {
+						continue;
+					}
+					$group->removeUser($user);
 				}
-				$this->groupManager->get($groupGID)?->removeUser($user);
+			}
+		} else {
+			$userGroupIds = $this->groupManager->getUserGroupIds($user);
+			foreach ($userGroupIds as $groupGID) {
+				if (!in_array($groupGID, array_column($syncGroups, 'gid'))) {
+					$this->groupManager->get($groupGID)?->removeUser($user);
+				}
 			}
 		}
 


lib/Service/ProvisioningService.php
```diff
@@ -762,13 +762,22 @@ public function provisionUserGroups(IUser $user, int $providerId, object $idToke
...
-		$userGroups = $this->groupManager->getUserGroupIds($user);
-		foreach ($userGroups as $groupGID) {
-			if (!in_array($groupGID, array_column($syncGroups, 'gid'))) {
-				if ($groupsWhitelistRegex && !preg_match($groupsWhitelistRegex, $groupGID)) {
-					continue;
+		if ($groupsWhitelistRegex) {
+			$userGroups = $this->groupManager->getUserGroups($user);
+			foreach ($userGroups as $group) {
+				if (!in_array($group->getGID(), array_column($syncGroups, 'gid'))) {
+					if (!preg_match($groupsWhitelistRegex, $group->getDisplayName())) {
```
julien-nc, line 769: Some group backends might return an empty group display name, right? If so, should we fallback to the GID?


```suggestion
					if (!preg_match($groupsWhitelistRegex, $group->getDisplayName() ?: $group->getGID())) {
```

julien-nc (commented):
Will this break the setup of instances configured with a regexp that matches group IDs?

lib/Service/ProvisioningService.php
OScaleSven: > If so, should we fallback to the GID?

In my opinion: yes. I have suggested that in my bug report already.

PR #1480, opened 2026-07-27
perf: Cache created user after creation

CarlSchwan:
## 🤖 AI (if applicable)

- [ ] The content of this PR was partly or fully generated using AI

### DIFF
--- a/lib/Controller/OcsApiController.php
+++ b/lib/Controller/OcsApiController.php
@@ -49,6 +49,9 @@ class OcsApiController extends OCSController {
 	): DataResponse {
 		$backendUser = $this->userMapper->getOrCreate($providerId, $userId);
 		$user = $this->userManager->get($backendUser->getUserId());
+		if ($user === null) {
+			throw new \RuntimeException('Unable to get user that was just created with userId=' . $backendUser->getUserId());
+		}
 
 		if ($displayName) {
 			if ($displayName !== $backendUser->getDisplayName()) {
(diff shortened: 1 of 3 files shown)


lib/User/Backend.php
```diff
@@ -126,7 +126,7 @@ public function getDisplayName($uid): string {
 		}
 		try {
 			$user = $this->userMapper->getUser($uid);
-			return $user->getDisplayName();
+			return $user->getDisplayName() ?? '';
```
julien-nc: This seems safe but why would we end up with a null display name?
The DB column default is `''`.
Do we need this precaution because we now potentially get a cached user that was freshly inserted and the DB entity is not loaded from DB so the display name is null?
If so, can we instead do `$user->setDisplayName('');` in `UserMapper::getOrCreate`?

Oh and also: We are missing the type declaration for the displayName column in the User DB entity class constructor.

julien-nc (approved):
Looks good. Optional concerns to address.

lib/User/Backend.php
CarlSchwan: The issue is that in lib/Db/User.php the default value is null, so if we don't re-fetch the object from the DB it stays null instead of the default '' from the DB. I refactored the code a bit now

I really don't like the Entity/Mapper api which is why I proposed a new one https://github.com/nextcloud/server/pull/56199 ;)

julien-nc (approved):

PR #1486, opened 2026-08-06
fix: don't set session user in getCurrentUserId

### LINKED ISSUE #1452: [Bug] setSessionUser() in getCurrentUserId() breaks loginWithExternalBackend() — no oc_authtoken created, AppPasswordController returns 403
### Summary

Since v8.7.0, bearer token authentication via `IApacheBackend` no longer creates a session token in `oc_authtoken`. As a result, any endpoint that relies on a valid session token after bearer authentication (e.g. Login Flow v2's `GET /ocs/v2.php/core/getapppassword`) returns **403**.

### Environment

- user_oidc: 8.10.1 (regression present since 8.7.0)
- Nextcloud: 30.x
- Provider configured with `check_bearer = 1` and `userinfo_bearer_validation = true`

### Root cause

PR #1376 introduced `setSessionUser()` calls inside `getCurrentUserId()` (the `IApacheBackend` callback). `IUserSession::setUser()` is called there, which sets `self::$_activeUser` in NC core's session before `loginWithExternalBackend()` completes.

NC core's `loginWithExternalBackend()` guards `createSessionToken()` with a check for an already-set active user — so once `setSessionUser()` runs inside `getCurrentUserId()`, no `oc_authtoken` row is ever written for the request.

When `AppPasswordController::getAppPassword()` then calls `LoginCredentialsStore::getLoginCredentials()`, `tokenProvider->getToken($sessionId)` throws `InvalidTokenException` because no row exists → 403.

### Logs

```
user_oidc GET /ocs/v2.php/core/getapppassword
Token validated with UserInfoValidator by provider: 1
admin_audit Login attempt: "<uid>"
core could not get login credentials because the token is invalid:
Token does not exist: token does not exist
→ status: 403
```

### Expected behaviour

Bearer token validation succeeds AND `loginWithExternalBackend()` completes its full flow (including `createSessionToken()`), so that session-dependent endpoints work correctly.

### Actual behaviour

`loginWithExternalBackend()` skips `createSessionToken()` because `setSessionUser()` already set the active user inside `getCurrentUserId()`. All endpoints that require a valid `oc_authtoken` (e.g. Login Flow v2 app-password endpoint) return 403.

### Suggested fix

Remove the `setSessionUser()` calls from `getCurrentUserId()` and instead register a listener on the `TokenValidatedEvent` that `user_oidc` already dispatches, calling `setSessionUser()` from there — after `loginWithExternalBackend()` has had the chance to complete.

### Version info

- user_oidc: 8.7.0 – 8.10.1 (regression introduced in 8.7.0 via PR #1376)
- Nextcloud: 30.x / 31.x
- Workaround: downgrade to v8.6.1

mosi-kha:
Fixes #1452

## Problem

`setSessionUser()`, added in #1376, is called from the first line of `getCurrentUserId()`. `loginWithApache()` guards its whole login block on the active user not already being set:

```php
$uid = $backend->getCurrentUserId();
if ($uid) {
    if (self::getUser() !== $uid) {     // <- guard
        ...
        $userSession->createSessionToken($request, $uid, $uid, $password);
        $userSession->createRememberMeToken($userSession->getUser());
        // token scopes, setupFS, post_login hook, UserLoggedInEvent
    }
    return true;
}
```

`IUserSession::setUser()` persists `user_id` into the session, so by the time control reaches the guard it's already satisfied and the entire block is skipped. No `oc_authtoken` row is ever created for that session.

## Impact

Not just app passwords (#1452's original symptom). `Session::validateSession()` looks up a token by session id on every request that reuses the session cookie; with no row to find, it calls `logout()`, which strips the cookie. A client that authenticates once with a bearer token and relies on the cookie for follow-up requests gets a 401 on the very next request.

Confirmed independently on master and `8.11.0-dev`: bearer request then cookie-only request goes `200` → `401` with the calls in place, `200` → `200` with them removed.

DAV also reaches this code, through its own `handleApacheAuth()` call site in `apps/dav/lib/Connector/Sabre/Auth.php`, separate from `OC::handleLogin()`. Confirmed `207` both before and after this change.

## Why a revert

`getCurrentUserId()` is called before `loginWithApache()`'s guard, and there's no hook in between — nothing the app does inside `getCurrentUserId()` can both satisfy the guard and leave a token row behind, since satisfying the guard *is* what skips the code that creates the row. A "complete the login inside the app" approach was considered and dropped: it only works by keeping the guard closed on purpose, which means maintaining a parallel login path here that has to keep tracking core's.

The `setupFS`/keygen cost this avoided is real, but it's a separate, filed issue — after this revert, every bearer request without a session cookie re-pays it, same as before #1376.

## Testing

`composer lint && composer cs:check && composer psalm` clean, no baseline changes.

No unit test added — `tests/unit/` has no existing coverage of `Backend.php`, and this is only observable end-to-end (bearer request → cookie reuse), which needs a running server, not a unit fixture.

### DIFF
--- a/lib/User/Backend.php
+++ b/lib/User/Backend.php
@@ -37,7 +37,6 @@ use OCP\ISession;
 use OCP\IURLGenerator;
 use OCP\IUser;
 use OCP\IUserManager;
-use OCP\IUserSession;
 use OCP\Server;
 use OCP\User\Backend\ABackend;
 use OCP\User\Backend\ICountUsersBackend;
@@ -369,12 +368,10 @@ class Backend extends ABackend implements IPasswordConfirmationBackend, IGetDisp
 							}
 
 							$this->session->set('last-password-confirm', $this->timeFactory->getTime() + 4 * 365 * 24 * 3600);
-							$this->setSessionUser($userId);
 							return $userId;
 						} elseif ($this->userExists($tokenUserId)) {
 							$this->checkFirstLogin($tokenUserId);
 							$this->session->set('last-password-confirm', $this->timeFactory->getTime() + 4 * 365 * 24 * 3600);
-							$this->setSessionUser($tokenUserId);
 							return $tokenUserId;
 						} else {
 							// check if the user exists locally
@@ -396,7 +393,6 @@ class Backend extends ABackend implements IPasswordConfirmationBackend, IGetDisp
 							}
 							$this->checkFirstLogin($tokenUserId);
 							$this->session->set('last-password-confirm', $this->timeFactory->getTime() + 4 * 365 * 24 * 3600);
-							$this->setSessionUser($tokenUserId);
 							return $tokenUserId;
 						}
 					}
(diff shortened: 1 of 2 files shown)


solracsf (approved):

solracsf:
Stand-by before merging, I'll try to write a regression tests here.

solracsf:
Pushed a regression test covering the three bearer return paths: each must resolve the user id without `IUserSession::setUser()` being called. Red with the `setSessionUser()` calls, green without.

The `psalm dev-master` failure is unrelated to this PR, server `95b69fd` removed `IIndex::hasColumnAtPosition()` on Aug 10 and `Version01022Date20221202161257.php` still calls it. Fix at https://github.com/nextcloud/user_oidc/pull/1493

pringelmann (approved):

PR #1493, opened 2026-08-11
fix(migration): stop using the removed IIndex::hasColumnAtPosition()

solracsf:
Nextcloud 35 reduced the schema API surface and dropped `hasColumnAtPosition()` from `OCP\DB\Schema\IIndex`, which fails the static-psalm-analysis dev-master job with four `UndefinedInterfaceMethod` errors in this migration.

Use `getColumns()` instead, which exists both on the new `IIndex` and on the Doctrine index objects still returned by Nextcloud 29 to 34, and compare the first column the way Doctrine did, unquoted and case insensitively, so the behaviour is unchanged on every supported server version.

### DIFF
--- a/lib/Migration/Version01022Date20221202161257.php
+++ b/lib/Migration/Version01022Date20221202161257.php
@@ -37,18 +37,19 @@ class Version01022Date20221202161257 extends SimpleMigrationStep {
 			$table = $schema->getTable('user_oidc_sessions');
 			$indexes = $table->getIndexes();
 			foreach ($indexes as $index) {
+				$columns = $index->getColumns();
 				// fix created_at index which is not unique
-				if ($index->isUnique() && $index->hasColumnAtPosition('created_at')) {
+				if ($index->isUnique() && $this->startsWithColumn($columns, 'created_at')) {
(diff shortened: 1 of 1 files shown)


CarlSchwan:
We can add hasColumnAtPosition back

solracsf:
Happy either way, though I don't think it needs to come back just for this.

There's also an argument for leaving it out. `Index::hasColumnAtPosition()` is [deprecated in doctrine/dbal 4.4](https://github.com/doctrine/dbal/blob/4.4.x/UPGRADE.md#deprecated-index-methods-properties-and-behavior) in favour of `getIndexedColumns()`, and it's built on top of `getUnquotedColumns()` and `getColumns()`, which are deprecated in the same release.

Putting it back on `IIndex` would tie a public `@since` API to something DBAL is already retiring.

Your call here.

CarlSchwan:
Done in https://github.com/nextcloud/server/pull/63152 this would obsolete your PR

solracsf:
Feel free to close it once merged :)

CarlSchwan:
Merged

PR #1498, opened 2026-08-24
feat(user_oidc): Add provider_id/sub columns and getByProviderAndSub()

CarlSchwan:
Store the provider and OIDC subject claim a user was provisioned from alongside the (possibly hashed) user_id, and look accounts up by this (provider_id, sub) pair first in getOrCreate(). This keeps a provider's own attribute drift (e.g. the mapped uid attribute changing) from forking the account into a second, empty one.

Existing rows are backfilled opportunistically on next login, since the original sub is not recoverable from a one-way hashed user_id.

## 🤖 AI (if applicable)

- [x] The content of this PR was partly or fully generated using AI

### DIFF
--- a/lib/Db/UserMapper.php
+++ b/lib/Db/UserMapper.php
@@ -152,7 +152,43 @@ class UserMapper extends QBMapper {
 		}
 	}
 
+	/**
+	 * @param non-empty-string $sub Sub of the user hashed if it exceed 256 characters
+	 * @throws \OCP\AppFramework\Db\DoesNotExistException
+	 * @throws \OCP\AppFramework\Db\MultipleObjectsReturnedException
+	 */
+	protected function getByProviderAndSub(int $providerId, string $sub): User {
+		$qb = $this->db->getQueryBuilder();
+		$qb->select('*')
+			->from($this->getTableName())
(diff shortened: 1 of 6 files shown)


lib/Db/UserMapper.php
```diff
@@ -152,22 +152,65 @@ public function userExists(string $uid): bool {
 		}
 	}
 
+	/**
+	 * @throws \OCP\AppFramework\Db\DoesNotExistException
+	 * @throws \OCP\AppFramework\Db\MultipleObjectsReturnedException
+	 */
```
come-nc, line 159: Should document that $sub is already hashed if needed before reaching this method.

lib/Migration/Version081100Date20260824120000.php
```diff
@@ -0,0 +1,48 @@
...
+				'length' => 256,
+			]);
+		}
+		if (!$table->hasIndex('user_oidc_prov_sub')) {
+			// unique: this is now the primary lookup key for an existing
+			// identity, not just an auxiliary index
+			$table->addUniqueIndex(['provider_id', 'sub'], 'user_oidc_prov_sub');
+		}
+
+		return $schema;
```
come-nc: Should return null if no changes

come-nc (approved):
Good apart from above nitpick

PR #1531, opened 2026-09-16
Allow JSON-Encoded group mapping

mwinkens:
Some upstream servers return each group object as a JSON-encoded string, e.g. "{\"gid\":\"group1\",\"displayName\":\"Group 1\"}". Decode these so the 'gid' and 'displayName' can be read correctly instead of being passed through verbatim.

Technically if a group send as a string looks like an encoded json object and would contain the gid key, this would be mapped wrong. However this is very unrealistic.

Also I know that this is technically not correct, but some identity proxies **including keycloak** deliver their group mapping like that. I'd argue that this allows nextcloud to handle a bug from outside instead of failing, and is therefore a useful feature.

## 🤖 AI (if applicable)

- [x] The content of this PR was partly or fully generated using AI (N/A)

I generated this PR with AI (opencode/Big Pickle Zen), but I **fully manually reviewed** it and **tested it in production**. I fully understand what this PR does.


CarlSchwan (commented):
Psaml failures are related

mwinkens:
> Psaml failures are related

I wrote a .patch file initially, and just wanted to merge this upstream - and then didn't copy the final `}` ... :laughing:

I didn't want to create a new branch in this project, I just created this PR over the github UI and it did create a branch for me (instead of forking and creating a branch). If this is accepted/declined/resolved, we can delete the branch again

susnux (approved):

CarlSchwan (approved):

PR #1390, opened 2026-03-26
refactor: modernize session management and token validation

solracsf:
This PR is a comprehensive audit and refactor, resolving bugs and performance problems identified through iterative code review. Refactor user session management and token validation logic for improved clarity and functionality. Introduce constants for session data and enhance error handling.

- Catch exceptions when instantiating bearer token validators, so a missing or broken DI entry no longer crashes the authentication chain.
- Catch exceptions during provisioning strategy resolution and execution, and log the failure instead of aborting login.
- Tighten bearer token parsing and basic UID validation before using the resolved user ID.
- Treat duplicate validations for the same (provider, userId) as valid, while still rejecting genuinely ambiguous matches across different providers or user IDs.
- Pre-instantiate validators outside provider loop

PR #1534, opened 2026-09-17
fix(backend): match the bearer auth-scheme case-insensitively

printminion-co:
### The problem

`Backend::getCurrentUserId()` guards the `Authorization` header with two case-sensitive checks for `bearer ` and `Bearer `, but the very next line strips the scheme with `/^bearer\s+/i`. The guard is therefore narrower than the code it protects: `BEARER token` and `Bearer<tab>token` are turned away before a validator ever runs, even though [RFC 9110 §11.1](https://www.rfc-editor.org/rfc/rfc9110.html#section-11.1) makes the auth-scheme case-insensitive and allows a horizontal tab as the separator.

Because this is the `IApacheBackend` entry point, an affected client fails on *every* API and WebDAV request, and it fails closed — the only trace is a `debug`-level `No Bearer token` line. I hit this on a real deployment.

| `Authorization:` | today | after |
|---|---|---|
| `Bearer t`, `bearer t` | ok | ok |
| `BEARER t`, `BeArEr t` | **rejected** | ok |
| `Bearer<tab>t` | **rejected** | ok |
| `Basic xyz` | rejected | rejected |

### The fix

One line: the guard now uses the same pattern as the strip that follows it, so you can see at a glance that the two agree.

```php
if (preg_match('/^bearer\s+/i', $headerToken) !== 1) {
```

(`!== 1` because `preg_match` returns `int<0,1>|false`.)

Nothing new is accepted: a missing header, a different scheme, a missing separator and empty credentials all still resolve to no user.

`isSessionActive()` is left alone on purpose — it only checks that the header is non-empty and does no scheme matching, so it has no case bug. Narrowing it would change how this backend interacts with other auth backends; that's a separate discussion.

### History

#1389 proposed the same fix the day after #1386 landed and was closed as superseded by #1390, which was split into #1418/#1419/#1420 and then closed — the case fix never came along. @solracsf is credited via `Co-authored-by`. Their version used `strtolower()`, which fixes the case half but still rejects a tab separator; hence the regex.

### AI assistance

Claude Code (`claude-opus-5`) assisted with the patch and the tests; the commit carries the `Assisted-by` trailer. I reviewed and ran everything myself and can defend every line.


solracsf (approved):
Makes sense but didn't test.

PR #1537, opened 2026-09-18
fix(login): block browser preload requests from consuming OIDC login state

technophile77:
## The problem

`LoginController::code()` is a `GET` that destructively consumes single-use server-side login state. When a browser speculatively prefetches or prerenders the callback URL, that background `GET` consumes the state before the user has clicked anything.

It is worse than losing the `oidc.*-<state>` session keys. `code()` calls `$this->userSession->completeLogin(...)`, which defaults to `$regenerateSessionId = true` and reaches `session_regenerate_id($deleteOldSession = true)` — so the winning request **deletes the session** that every other in-flight callback for that flow depends on. The user's real navigation then arrives holding a cookie for a session that no longer exists, `$storedState` is `null`, and the login is refused.

Because the IdP → RP callback is a cross-site navigation, Chrome prefetches it in a *partitioned* network context, so the `Set-Cookie` from the winning preload never reaches the user's real cookie jar. The login succeeds inside a context that is then discarded, and the user is shown "Access forbidden" seconds after authorising the request.

This is the same defect pattern as rotating a per-request CSRF token on `GET`: the preloader absorbs the single-use token and the user's real tab is left holding something the server no longer recognises.

## Evidence

From a production instance, callback requests grouped by `state` over one day:

| Pattern per state | Flows | Meaning |
|---|---|---|
| `303` only | 5 | clean login |
| `303` plus 1-4 × `403` | 5 | succeeded, duplicate callbacks errored |
| `403` only | 5 | **never succeeded** |
| `403 403` | 2 | **never succeeded** |

18 flows, 19 × `403` against 12 × `303`. Seven flows (39%) produced no success at all, and users recovered only by retrying. Duplicates skew heavily mobile.

## The change

Detect the standard browser speculative-loading headers and decline the request before any state is touched, at all three doors that mint or consume it:

- `LoginController::code()` — as the first statement, ahead of the `isLoggedIn()` early return, which itself calls `cleanupSessionState()`
- `LoginController::login()` — as the first statement, so a preloaded login never produces an `/authorize` hit (this is what stops the IdP minting a second authorization code for the same flow)
- `Application::registerRedirect()` — a speculative preload of `/login` currently *passes* the existing `isTopLevelHtmlNavigation()` check, so it is auto-redirected to the IdP today

Per the Speculation Rules specification, a non-success status makes the browser **discard the speculation entirely**, so the prerender is never activated and the user's real click issues a fresh, normal request that completes the flow.

### Notes for review

- **`Sec-Purpose` is parsed as a token list, not compared with `===`.** Chromium sends `Sec-Purpose: prefetch;prerender` for a prerendered navigation, so an equality check silently never matches a prerender. The added tests pin this specifically.
- **The classification is never used to suppress rate limiting or brute-force protection.** These headers are client-controlled and trivially spoofed; skipping state consumption grants a spoofer nothing, since it only denies them their own login. Wiring them to throttling would be a one-header bypass, and this PR deliberately does not do that.
- Logging is at `debug` only.
- Safari sends none of these headers, so it is not covered by this change and falls back to existing behaviour.

## Tests

`RequestClassificationService` was chosen as the home for the predicate precisely because it is unit-testable — `LoginController` has 23 constructor dependencies and no existing coverage. The added data provider covers the absent header, an empty value, `prefetch`, `prerender`, `prefetch;prerender`, `prefetch; prerender` with a space, mixed case, comma-separated lists, each legacy header, and negative cases including a matching token appearing only in an unrelated header.

## Depends on #1536

This is stacked on #1536, which fixes the arithmetic that made this failure report itself as an expired state (with `$sessionTimestamp === null`, `$currentTimestamp - $sessionTimestamp > 300` reduces to `$now - 0 > 300`, unconditionally true). The lookup added here lives inside the branch that PR introduces. Happy to rebase onto `main` once it lands.


technophile77:
Withdrawing this — the premise turned out to be wrong, and I'd rather say so than leave it open.

The theory was that browser speculative loading (prefetch/prerender) was consuming single-use OIDC login state before the user's real navigation arrived. I deployed this change to a production instance together with additional logging to confirm it. It didn't hold up.

The instrumentation added `Sec-Purpose`, `Purpose`, `X-Purpose` and the `Sec-Fetch-*` headers to the access log, plus the real client IP. That last part mattered more than expected: the origin sits behind a CDN and had been logging edge addresses rather than client ones, which had made three duplicate callbacks look like three separate clients when they were in fact three CDN edges hitting the same flow.

With that in place, across real user traffic **the guard added by this PR never rejected a single request**. Every duplicate callback was a genuine top-level navigation — `Sec-Fetch-Mode: navigate`, no `Sec-Purpose` header present. No speculative request was observed reaching the endpoint at all.

The actual cause was outside Nextcloud: our identity provider's OAuth consent screen could be submitted repeatedly, and each submission minted a fresh authorization code under the same `state`. The provider's own access log showed a single `GET … 200` rendering the consent form, followed by four `POST … 302` submissions one second apart. Nextcloud then received several callbacks for one login flow — the first consumed it, and the rest correctly rendered "Access forbidden". That is being fixed on the provider side, where it belongs.

So there's nothing left of the original premise to rework, and leaving this open would imply a problem in user_oidc that I have no evidence for. Closing.

